PR #3, opened 2020-04-07
Mono: Add support for iOS

akien-mga:
It seems to work fine with the (intermediate) container build that I get with https://github.com/godotengine/build-containers/pull/47 (without cross-compiling toolchain).

BTW, how does the AOT toolchain intervene in the build/packaging of the templates? Or is it something that should be bundled alongside the editor builds for all desktop platforms?

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -177,8 +177,8 @@ if [ "${build_classical}" == "1" ]; then
   rm -rf ios_xcode
   cp -r git/misc/dist/ios_xcode ios_xcode
   for suffix in "" "_arkit_module" "_camera_module"; do
-    cp out/ios/libgodot${suffix}.iphone.opt.fat ios_xcode/libgodot${suffix}.iphone.release.fat.a
-    cp out/ios/libgodot${suffix}.iphone.opt.debug.fat ios_xcode/libgodot${suffix}.iphone.debug.fat.a
+    cp out/ios/templates/libgodot${suffix}.iphone.opt.fat ios_xcode/libgodot${suffix}.iphone.release.fat.a
+    cp out/ios/templates/libgodot${suffix}.iphone.opt.debug.fat ios_xcode/libgodot${suffix}.iphone.debug.fat.a
   done
   chmod +x ios_xcode/libgodot*.iphone.*
   cd ios_xcode
@@ -234,6 +234,7 @@ if [ "${build_mono}" == "1" ]; then
   mkdir -p ${binbasename}_64
   cp out/linux/x64/tools-mono/godot.x11.opt.tools.64.mono ${binbasename}_64/${binbasename}.64
   cp -rp out/linux/x64/tools-mono/GodotSharp ${binbasename}_64/
+  cp -rp out/aot-compilers ${binbasename}_64/GodotSharp/Tools/
   zip -r -q -9 "${reldir_mono}/${binbasename}_64.zip" ${binbasename}_64
   rm -rf ${binbasename}_64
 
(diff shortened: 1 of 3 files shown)


neikeq:
>BTW, how does the AOT toolchain intervene in the build/packaging of the templates? Or is it something that should be bundled alongside the editor builds for all desktop platforms?

https://docs.godotengine.org/en/latest/development/compiling/compiling_with_mono.html#aot-cross-compilers

I just noticed the path mentioned there is wrong. It should be: `GodotSharp/Tools/aot-compilers/<platform>-<arch>` (`platform` as in what you pass to scons and `arch` must be the one Godot adds to features).

Example for iOS:

```bash
mkdir -p bin/GodotSharp/Tools/aot-compilers/iphone-arm64
cp /root/mono-installs/ios-arm64-release/aarch64-apple-darwin-mono-sgen bin/GodotSharp/Tools/aot-compilers/iphone-arm64
```
I don't remember if `aarch64-apple-darwin-mono-sgen` is the correct name of the executable though.

akien-mga:
I pushed an update that should complete the packaging of iOS Mono builds with the cross-compiler that you provided, which I included directly in the build container (as `/root/aot-compilers/iphone-arm64/aarch64-apple-darwin-mono-sgen`).

akien-mga:
Here's the output of those updated build scripts (some files still uploading): https://downloads.tuxfamily.org/godotengine/testing/3.2.2-mono_ios_build1/

akien-mga:
Woops... I'm good for repacking:
```
GodotSharp/Tools/iphone-arm64/aarch64-apple-darwin-mono-sgen
```

build-release.sh
```diff
@@ -234,13 +234,15 @@ if [ "${build_mono}" == "1" ]; then
   mkdir -p ${binbasename}_64
   cp out/linux/x64/tools-mono/godot.x11.opt.tools.64.mono ${binbasename}_64/${binbasename}.64
   cp -rp out/linux/x64/tools-mono/GodotSharp ${binbasename}_64/
+  cp -rp out/aot-compilers ${binbasename}_64/GodotSharp/Tools/
```
akien-mga, line 237: As discussed on Discord, the ios cross-compiler is only usable from macOS, so it should only be included with the macOS editor binary, not Windows/Linux. I'll edit this.

build-release.sh
```diff
@@ -359,7 +365,22 @@ if [ "${build_mono}" == "1" ]; then
...
-  # Not supported yet.
+  rm -rf ios_xcode
+  cp -r git/misc/dist/ios_xcode ios_xcode
+  for suffix in "" "_arkit_module" "_camera_module"; do
+    cp out/ios/templates-mono/libgodot${suffix}.iphone.opt.fat ios_xcode/libgodot${suffix}.iphone.release.fat.a
+    cp out/ios/templates-mono/libgodot${suffix}.iphone.opt.debug.fat ios_xcode/libgodot${suffix}.iphone.debug.fat.a
+  done
+  chmod +x ios_xcode/libgodot*.iphone.*
+  cd ios_xcode
+  zip -q -9 -r "${templatesdir}/iphone.zip" *
```
neikeq, line 376: I think this is why `iphone.zip` was missing from the test build:

```suggestion
  zip -q -9 -r "${templatesdir_mono}/iphone.zip" *
```

akien-mga: Thanks, fixed.

akien-mga:
I added the missing `chmod +x` for the AOT compiler(s), so this should now be good to go.

Let's see how it fares tomorrow with a 3.2.2-beta1 build.

PR #10, opened 2020-07-08
OSX: Build universal binaries with ARM64

akien-mga:
Untested for now, but preparing the ground for building universal binaries with ARM64 support for macOS.

Requirements to finish this:

- 1. [x] osxcross should first get support for the ARM64 target ([upstream issue](https://github.com/tpoechtrager/osxcross/issues/227)).
- 2. [x] [Dockerfile.xcode](https://github.com/godotengine/build-containers/blob/master/Dockerfile.xcode) and [Dockerfile.osx](https://github.com/godotengine/build-containers/blob/master/Dockerfile.osx) will need to be updated to build with Xcode 12 and the latest osxcross version when it supports ARM64. (Done with https://github.com/godotengine/build-containers/pull/60.)
- 3. [ ] [Dockerfile.osx](https://github.com/godotengine/build-containers/blob/master/Dockerfile.osx) should be updated to build Mono for ARM64.
   - [x] This depends on Mono actually getting [support for ARM64 merged](https://github.com/mono/mono/pull/20014) in the first place, 
   - [x] and also implies that we'd have to build against a fairly cutting-edge Mono version to get this support, as I guess they might not backport it to earlier branches. (Done with https://github.com/godotengine/build-containers/pull/58 and https://github.com/godotengine/build-containers/pull/59.)

Sooo... it's not for today ;)

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -136,11 +136,11 @@ if [ "${build_classical}" == "1" ]; then
   ## OSX (Classical) ##
 
   # Editor
-  binname="${godot_basename}_osx.64"
+  binname="${godot_basename}_osx.universal"
   rm -rf Godot.app
   cp -r git/misc/dist/osx_tools.app Godot.app
   mkdir -p Godot.app/Contents/MacOS
-  cp out/macosx/x64/tools/godot.osx.opt.tools.64 Godot.app/Contents/MacOS/Godot
+  cp out/macosx/tools/godot.osx.opt.tools.universal Godot.app/Contents/MacOS/Godot
   chmod +x Godot.app/Contents/MacOS/Godot
   zip -q -9 -r "${reldir}/${binname}.zip" Godot.app
   rm -rf Godot.app
@@ -150,8 +150,8 @@ if [ "${build_classical}" == "1" ]; then
   cp -r git/misc/dist/osx_template.app .
   mkdir -p osx_template.app/Contents/MacOS
 
-  cp out/macosx/x64/templates/godot.osx.opt.64 osx_template.app/Contents/MacOS/godot_osx_release.64
-  cp out/macosx/x64/templates/godot.osx.opt.debug.64 osx_template.app/Contents/MacOS/godot_osx_debug.64
+  cp out/macosx/templates/godot.osx.opt.universal osx_template.app/Contents/MacOS/godot_osx_release.64
+  cp out/macosx/templates/godot.osx.opt.debug.universal osx_template.app/Contents/MacOS/godot_osx_debug.64
   chmod +x osx_template.app/Contents/MacOS/godot_osx*
   zip -q -9 -r "${templatesdir}/osx.zip" osx_template.app
   rm -rf osx_template.app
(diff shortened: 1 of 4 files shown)


build-release.sh
```diff
@@ -324,18 +324,18 @@ if [ "${build_mono}" == "1" ]; then
...
-  cp out/macosx/x64/tools-mono/godot.osx.opt.tools.64.mono Godot_mono.app/Contents/MacOS/Godot
+  cp out/macosx/tools-mono/godot.osx.opt.tools.universal.mono Godot_mono.app/Contents/MacOS/Godot
   mkdir -p Godot_mono.app/Contents/{Frameworks,Resources}
   mkdir -p Godot_mono.app/Contents/{Frameworks,Resources}/GodotSharp
   mkdir -p Godot_mono.app/Contents/{Frameworks,Resources}/GodotSharp/Mono
-  cp -rp out/macosx/x64/tools-mono/GodotSharp/Api Godot_mono.app/Contents/Frameworks/GodotSharp
-  cp -rp out/macosx/x64/tools-mono/GodotSharp/Mono/lib Godot_mono.app/Contents/Frameworks/GodotSharp/Mono
-  cp -rp out/macosx/x64/tools-mono/GodotSharp/Tools Godot_mono.app/Contents/Frameworks/GodotSharp
-  cp -rp out/macosx/x64/tools-mono/GodotSharp/Mono/etc Godot_mono.app/Contents/Resources/GodotSharp/Mono
+  cp -rp out/macosx/tools-mono/GodotSharp/Api Godot_mono.app/Contents/Frameworks/GodotSharp
```
bruvzg, line 335: Since official releases are not signed, it's currently not relevant, but I do not think placing non-native binaries in the `Frameworks` is correct, probably `dll`s should go to `Resources` instead.

MintStudios:
How is the progress for this going?

akien-mga:
I need to give this a new try now that OSXCross has been updated with preliminary support for ARM64.

I have Xcode 12.0.1 available on the buildsystem, but I'm not sure yet if it includes the relevant SDK (at least what I extracted from it so far is the MacOSX10.15 SDK).

bruvzg:
> I have Xcode 12.0.1

I do not think it is. 12.0 had a separate beta for Universal, and it was not included into release. Probably ARM SDK is included in 12.2 RC only (SDK version and should be `MacOSX11.0.sdk`).

akien-mga:
Makes sense, thanks. I'll wait for 12.2 to be released as stable, I don't specifically need/want to be an early adopter and run into testing issues for Apple software... especially when using a non-standard toolchain like OSXCross on top of it.

bruvzg:
>  I'll wait for 12.2 to be released as stable

Xcode 12.2 stable is released, and building ARM64 Godot (both `master` and `3.2` branches) without any issues. I have checked `release-debug` tool builds without Mono only.

On Silicon Macs all executables must be signed, but any signature is fine (Gatekeeper will be complaining if it's not Apple approved signature and the app is not notarized, in the same way it does on macOS 10.15).

Xcode linker is adding ad-hoc signatures automatically, and it's preserved when `lipo` is used to create universal binary, OSXCross should do the same, but I not sure how automated it is.

akien-mga:
I've now used the containers from https://github.com/godotengine/build-containers/pull/60 and the build scripts from this PR to build 3.2.4-beta2 for macOS arm64:
https://downloads.tuxfamily.org/godotengine/3.2.4/beta2/Godot_v3.2.4-beta2_osx.universal.zip

This is only the standard build, I haven't looked into building Mono for arm64 yet. I'll do that in a separate PR (and maybe not right now as it seems @bruvzg had troubles on real hardware too, so it might be worth waiting for first issues to be ironed out at Mono upstream).

build-release.sh
akien-mga: I'm not familiar with the implication of this change (especially whether the binary would still find them properly). Might need to be checked with @neikeq (who should get a M1 device in coming weeks too).

MintStudios:
Thanks!

monokrome:
Thank you all for being so amazing

PR #12, opened 2020-08-27
JavaScript: Enable lto.

Faless:


akien-mga:
Thanks!

akien-mga:
As discussed on IRC, I'm temporarily disabling LTO again as it doesn't work with our current 1.38.47-upstream emscripten version (279085ccce41baf845af7e8f47713a558bf08225).

We need Emscripten 1.39.7+ to properly support LTO it seems (not tested that version specifically yet, but that's when https://github.com/emscripten-core/emscripten/issues/10066 was fixed).

That means upgrading to Mono 6.12.x or newer: https://github.com/godotengine/build-containers/pull/44#issuecomment-609717400

PR #13, opened 2020-11-22
Fixing Mono export templates and editor

sjml:
This pairs with godotengine/godot#43768, which modifies the editor's export process to produce `.app`s that are able to be codesigned. With these changes, the official builds would be able to be signed, but more importantly, the pre-made templates would be ready to go.

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -330,10 +330,12 @@ if [ "${build_mono}" == "1" ]; then
   mkdir -p Godot_mono.app/Contents/MacOS
   cp out/macosx/tools-mono/godot.osx.opt.tools.universal.mono Godot_mono.app/Contents/MacOS/Godot
   mkdir -p Godot_mono.app/Contents/{Frameworks,Resources}
-  mkdir -p Godot_mono.app/Contents/{Frameworks,Resources}/GodotSharp
-  cp -rp out/macosx/tools-mono/GodotSharp/Api Godot_mono.app/Contents/Frameworks/GodotSharp
-  cp -rp out/macosx/tools-mono/GodotSharp/Tools Godot_mono.app/Contents/Frameworks/GodotSharp
-  cp -rp out/macosx/tools-mono/GodotSharp/Mono Godot_mono.app/Contents/Resources/GodotSharp
+  mkdir -p Godot_mono.app/Contents/Resources/GodotSharp
+  mkdir -p Godot_mono.app/Contents/Frameworks/Mono
+  cp -rp out/macosx/tools-mono/GodotSharp Godot_mono.app/Contents/Resources
+  mv Godot_mono.app/Contents/Resources/GodotSharp/Mono/lib/*.dylib Godot_mono.app/Contents/Frameworks/Mono
+  # cp /usr/local/lib/libMoltenVK.dylib Godot_mono.app/Contents/Frameworks
+  sed -i '' -e 's/$mono_libdir/$mono_libdir\/..\/..\/..\/..\/Frameworks\/Mono/g' Godot_mono.app/Contents/Resources/GodotSharp/Mono/etc/mono/config
 
   cp -rp out/aot-compilers Godot_mono.app/Contents/Frameworks/GodotSharp/Tools/
   chmod +x Godot_mono.app/Contents/MacOS/Godot
@@ -347,7 +349,15 @@ if [ "${build_mono}" == "1" ]; then
 
   cp out/macosx/templates-mono/godot.osx.opt.debug.universal.mono osx_template.app/Contents/MacOS/godot_osx_debug.64
   cp out/macosx/templates-mono/godot.osx.opt.universal.mono osx_template.app/Contents/MacOS/godot_osx_release.64
-  cp -rp out/macosx/templates-mono/data.mono.osx.universal.* osx_template.app/Contents/Resources/
+
+  mkdir -p osx_template.app/Contents/Frameworks/data.mono.osx.64.frameworks.{release,release_debug}/lib
+  mkdir -p osx_template.app/Contents/Resources/data.mono.osx.64.resources.{release,release_debug}
+  cp -rp out/templates-mono/data.mono.osx.64.release/Mono/lib osx_template.app/Contents/Frameworks/data.mono.osx.64.frameworks.release
+  cp -rp out/templates-mono/data.mono.osx.64.release_debug/Mono/lib osx_template.app/Contents/Frameworks/data.mono.osx.64.frameworks.release_debug
+  cp -rp out/templates-mono/data.mono.osx.64.release/Mono/etc osx_template.app/Contents/Resources/data.mono.osx.64.resources.release
+  cp -rp out/templates-mono/data.mono.osx.64.release_debug/Mono/etc osx_template.app/Contents/Resources/data.mono.osx.64.resources.release_debug
+  # cp /usr/local/lib/libMoltenVK.dylib osx_template.app/Contents/Frameworks
+
   chmod +x osx_template.app/Contents/MacOS/godot_osx*
   zip -q -9 -r "${templatesdir_mono}/osx.zip" osx_template.app
   rm -rf osx_template.app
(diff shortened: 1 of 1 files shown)


build-release.sh
```diff
@@ -331,11 +331,11 @@ if [ "${build_mono}" == "1" ]; then
   cp out/macosx/tools-mono/godot.osx.opt.tools.universal.mono Godot_mono.app/Contents/MacOS/Godot
   mkdir -p Godot_mono.app/Contents/{Frameworks,Resources}
   mkdir -p Godot_mono.app/Contents/{Frameworks,Resources}/GodotSharp
-  mkdir -p Godot_mono.app/Contents/{Frameworks,Resources}/GodotSharp/Mono
+  mkdir -p Godot_mono.app/Contents/Resources/GodotSharp/Mono
   cp -rp out/macosx/tools-mono/GodotSharp/Api Godot_mono.app/Contents/Frameworks/GodotSharp
-  cp -rp out/macosx/tools-mono/GodotSharp/Mono/lib Godot_mono.app/Contents/Frameworks/GodotSharp/Mono
   cp -rp out/macosx/tools-mono/GodotSharp/Tools Godot_mono.app/Contents/Frameworks/GodotSharp
-  cp -rp out/macosx/tools-mono/GodotSharp/Mono/etc Godot_mono.app/Contents/Resources/GodotSharp/Mono
+  cp -rp out/macosx/tools-mono/GodotSharp/Mono Godot_mono.app/Contents/Resources/GodotSharp/Mono
```
akien-mga, line 337: IINM this will end up as `Godot_mono.app/Contents/Resources/GodotSharp/Mono/Mono`, is that intended? If not, the `mkdir` should be removed and then the destination should be the `GodotSharp/` folder.

sjml: Yes, my bad. That's a copy-paste casualty. (I don't have the means of reproducing the entire Godot build setup here, unfortunately, so having to kind of shunt my own local scripts into it.) Will fix. Thanks!

sjml:
Updated this to match the layouts for the engine pull request (godotengine/godot#43768). Marking this as a draft until that one is accepted, since it's irrelevant and actually harmful until then. 

One question for you, though @akien-mga -- how do you feel about the use of `sed` to do some post-processing on a file? (Still trying to see if there's a way to make Mono happier with a split `lib` directory, but I'm not too optimistic.) It's pretty straightforward, but I notice that the rest of this script doesn't do any thing like that so just wanted to flag it. Happy to edit the file in some other way if you'd prefer.

akien-mga:
> One question for you, though @akien-mga -- how do you feel about the use of sed to do some post-processing on a file? (Still trying to see if there's a way to make Mono happier with a split lib directory, but I'm not too optimistic.) It's pretty straightforward, but I notice that the rest of this script doesn't do any thing like that so just wanted to flag it. Happy to edit the file in some other way if you'd prefer.

The packaging process should not become too convoluted, end users should also be able to produce their own Mono builds manually if they wish to. If something convoluted has to be done, it's best if it was done by SCons itself.

sjml:
> If something convoluted has to be done, it's best if it was done by SCons itself.

That makes sense to me. The trouble is that packaging the Mac `.app` directory currently happens in the `build-release` script, after SCons is out of the picture. Would you like to see this packaging get moved to SCons so the `build-release` script only has to fetch a more final artifact and zip it up? (I'm wary of introducing this idea since addressing it it likely entails my learning SCons, but if that's the way it should be done...)

akien-mga:
Having the bundle auto-generated could be nice, but that's outside the scope of the fixes you're doing I think, and not necessarily convenient (especially now that we need to build for two arches again before making the bundle).

What kind of `sed` changes do you need to do?

sjml:
> What kind of `sed` changes do you need to do?

You can see at line 338 here: https://github.com/godotengine/godot-build-scripts/pull/13/files#diff-56d440968971bab627980eb075620a7939bc1d91395d9f6c5d4814277dd3ff47

It's rewriting some library paths in the Mono config so the executable can find them based on the `lib` path that it has to set to find all the DLLs. 

(I can remove those commented out copying of the `libMoltenVK.dylib` there, too; they are more there as a reminder to me that it's needed to have a valid build.)

akien-mga:
Hm that's a bit of an annoying extra step to the early fairly complex instructions in https://docs.godotengine.org/en/latest/development/compiling/compiling_with_mono.html

WDYT @neikeq?

sjml:
> that's a bit of an annoying extra step

Very much agreed. You can see the evolution of the conversation on the engine pull request. I'm still trying to see if there's a better way to get the desired directory layout *and* have Mono functional, but since it doesn't expect its `lib` directory to be split up, something hacky like this might be needed.

bruvzg (approved):

build-release.sh
```diff
@@ -330,12 +330,13 @@ if [ "${build_mono}" == "1" ]; then
...
-  cp -rp out/macosx/tools-mono/GodotSharp/Api Godot_mono.app/Contents/Frameworks/GodotSharp
-  cp -rp out/macosx/tools-mono/GodotSharp/Mono/lib Godot_mono.app/Contents/Frameworks/GodotSharp/Mono
-  cp -rp out/macosx/tools-mono/GodotSharp/Tools Godot_mono.app/Contents/Frameworks/GodotSharp
-  cp -rp out/macosx/tools-mono/GodotSharp/Mono/etc Godot_mono.app/Contents/Resources/GodotSharp/Mono
+  mkdir -p Godot_mono.app/Contents/Resources/GodotSharp
+  mkdir -p Godot_mono.app/Contents/Frameworks/Mono
+  cp -rp out/macosx/tools-mono/GodotSharp Godot_mono.app/Contents/Resources
+  mv Godot_mono.app/Contents/Resources/GodotSharp/Mono/lib/*.dylib Godot_mono.app/Contents/Frameworks/Mono
+  # cp /usr/local/lib/libMoltenVK.dylib Godot_mono.app/Contents/Frameworks
+  sed -i -e 's/$mono_libdir/$mono_libdir\/..\/..\/..\/..\/Frameworks\/Mono/g' Godot_mono.app/Contents/Resources/GodotSharp/Mono/etc/mono/config
```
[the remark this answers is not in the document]
sjml: I think it will also do the same thing with macOS sed. Usually I've done a check and made a `$SED` variable, but was trying to avoid having yet-another-complication added here, since these scripts are run in Linux containers. If we want to make sure it's more portable, though, that's an easy bit of logic to add.

akien-mga:
With @neikeq's changes added to #43768 this should now be simpler, as we get this structure:

### Tools

```
$ tree -d tools-mono/
tools-mono/
├── GodotSharp
│   ├── Api
│   │   ├── Debug
│   │   └── Release
│   └── Tools
├── GodotSharp.frameworks
│   └── Mono
│       └── lib
└── GodotSharp.resources
    ├── Mono
    │   ├── etc
    │   │   └── mono
    │   │       ├── 2.0
    │   │       │   └── Browsers
    │   │       ├── 4.0
    │   │       │   └── Browsers
    │   │       ├── 4.5
    │   │       │   └── Browsers
    │   │       └── mconfig
    │   └── lib
    │       └── mono
    │           └── 4.5
    │               └── Facades
    └── Tools
```

### Templates

```
$ tree -d templates-mono/
templates-mono/
├── data.mono.osx.64.frameworks.release
│   └── lib
├── data.mono.osx.64.frameworks.release_debug
│   └── lib
├── data.mono.osx.64.resources.release
│   └── etc
│       └── mono
│           ├── 2.0
│           │   └── Browsers
│           ├── 4.0
│           │   └── Browsers
│           ├── 4.5
│           │   └── Browsers
│           └── mconfig
└── data.mono.osx.64.resources.release_debug
    └── etc
        └── mono
            ├── 2.0
            │   └── Browsers
            ├── 4.0
            │   └── Browsers
            ├── 4.5
            │   └── Browsers
            └── mconfig
```

akien-mga:
Superseded by #24.

PR #14, opened 2020-12-10
JS: Build with threads and gdnative where supported + re-enable LTO

akien-mga:
Classical builds now use Emscripten 2.0.10 and have threads and gdnative templates.
Mono builds still use Emscripten 1.39.9 for compatibility with Mono 6.12.x, so they
can only have threads support (GDNative requires 2.0.10+).


akien-mga:
Seems to work fine. Requires updated JS container as of https://github.com/godotengine/build-containers/pull/62.

PR #15, opened 2021-01-24
Update build.sh

qjyqjyqjyqjy:
-  export EMSDK_CLASSICAL=2.0.10
+ export EMSDK_CLASSICAL=2.0.11


akien-mga:
Thanks! Fixed with #17 instead as those variables are now defined in the container itself.

PR #20, opened 2021-02-12
[HTML5] Add tools builds to release scripts.

Faless:
Will unpack the zip, compress each file with brotli, and place them into the `web` folder (to be uploaded to the web editor host server).

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -60,6 +60,7 @@ if [ -z "${godot_version}" -o -z "${templates_version}" ]; then
 fi
 
 export basedir=$(pwd)
+export webdir="${basedir}/web/${godot_version}"
 export reldir="${basedir}/releases/${godot_version}"
 export reldir_mono="${reldir}/mono"
 export tmpdir="${basedir}/tmp"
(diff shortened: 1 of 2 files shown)


build-release.sh
```diff
@@ -73,6 +74,7 @@ export godot_basename="Godot_v${godot_version}"
 rm -rf ${reldir}
 rm -rf ${tmpdir}
 
+mkdir -p ${webdir}
```
akien-mga, line 77: Should also be `rm -rf`'ed above.

build-release.sh
```diff
@@ -172,6 +174,13 @@ if [ "${build_classical}" == "1" ]; then
 
   ## Javascript (Classical) ##
 
+  # Web (editor)
```
akien-mga, line 177:
```suggestion
  # Editor
```

build-release.sh
```diff
@@ -172,6 +174,13 @@ if [ "${build_classical}" == "1" ]; then
...
+  rm -rf ${tmpdir}/webtmp/
+  mkdir -p ${tmpdir}/webtmp/
```
akien-mga, line 179: I guess you could unzip in `${webdir}/` directly and not need `${tmpdir}/webtmp/`?

akien-mga (approved):

akien-mga:
Thanks!

PR #21, opened 2021-02-12
[WIP] tvOS build support

naithar:
Allows to build tvOS binaries.
Depends on https://github.com/godotengine/godot/pull/45829


build-tvos/build.sh
```diff
@@ -0,0 +1,47 @@
+#!/bin/bash
+
+set -e
+
+# Config
+
+export BUILD_NAME=official
```
Xrayez, line 7: This will have to removed due to #25.

```suggestion

```

naithar: Done, thanks!

PR #22, opened 2021-02-22
Use new production=yes option

akien-mga:
Equivalent to debug_symbols=no use_lto=yes use_static_cpp=yes.
We keep LTO disabled for iOS as users need to relink on deploy, and that's very slow
and memory hungry with LTO.

### DIFF
--- a/build-ios/build.sh
+++ b/build-ios/build.sh
@@ -6,7 +6,9 @@ set -e
 
 export BUILD_NAME=official
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
-export OPTIONS="debug_symbols=no"
+# Keep LTO disabled for iOS - it works but it makes linking apps on deploy very slow,
+# which is seen as a regression in the current workflow.
+export OPTIONS="production=yes use_lto=no"
 export OPTIONS_MONO="module_mono_enabled=yes mono_static=yes"
 export TERM=xterm
 
--- a/build-android/build.sh
+++ b/build-android/build.sh
@@ -6,7 +6,7 @@ set -e
 
 export BUILD_NAME=official
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
-export OPTIONS="debug_symbols=no"
+export OPTIONS="production=yes"
 export OPTIONS_MONO="module_mono_enabled=yes mono_static=no"
 export TERM=xterm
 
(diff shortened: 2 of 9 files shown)


build-ios/build.sh
```diff
@@ -6,7 +6,9 @@ set -e
 
 export BUILD_NAME=official
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
-export OPTIONS="debug_symbols=no"
+# Keep LTO disabled for iOS - it works but it makes linking apps on deploy very slow,
+# which is seen as a regression in the current workflow.
+export OPTIONS="production=yes use_lto=no"
```
akien-mga, line 11: @naithar 3.2.4 RC 2 used this with `OPTIONS="production=yes"` only.
I guess this change should fix https://github.com/godotengine/godot/issues/46064.

bruvzg: What about `-flto=thin`, it should be supported and work much faster.

akien-mga: We could try it, though that requires changes to `platform/iphone/detect.py`. Might also be worth reworking how we pass LTO options as e.g. for `-flto=thin` on Linux you need to pass `use_lto=yes use_lld=yes use_thinlto=yes`.

naithar: > What about `-flto=thin`, it should be supported and work much faster.

It was 3x faster for tvOS builds https://github.com/godotengine/godot-build-scripts/pull/21, so we can try using it. I can still provide macOS non-lto builds just in case.

@akien-mga should I also change tvOS build script to support `production` flag?

akien-mga: > should I also change tvOS build script to support `production` flag?

Yeah, sounds good.

bruvzg: > What about -flto=thin, it should be supported and work much faster.

And it's using much less memory (under 900 MB), waiting for a few minutes for release builds seems pretty reasonable.

bruvzg: Also, LTO flags are completely missing from macOS `detect.py`.

akien-mga: Yeah I think we should overhaul the LTO flags to make sure they're consistent across all platforms and use the most optimal configuration by default (so `thin` on Clang/LLD).

akien-mga: I opened https://github.com/godotengine/godot/issues/46317 to continue the discussion on LTO for iOS and macOS.

akien-mga: I had a small doubt whether this override would work... and it turns out it didn't :)
Needs some fixes to how we handle overriding options that would get default values from `dev=yes` and `production=yes`: https://github.com/godotengine/godot/pull/46365

PR #23, opened 2021-02-28
macOS editor entitlements

naithar:
Fixes https://github.com/godotengine/godot/issues/46450 as well as possible crash related to microphone access permission.

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -152,10 +152,11 @@ if [ "${build_classical}" == "1" ]; then
     osx_tmpdir=$(ssh "${OSX_HOST}" "mktemp -d")
     
     scp "${reldir}/${binname}.zip" "${OSX_HOST}:${osx_tmpdir}"
+    scp "${basedir}/build-macosx/editor.entitlements" "${OSX_HOST}:${osx_tmpdir}"
     ssh "${OSX_HOST}" "
               cd ${osx_tmpdir} && \
               unzip ${binname}.zip &&\
-              codesign --timestamp --options=runtime -s ${OSX_KEY_ID} -v Godot.app/Contents/MacOS/Godot && \
+              codesign --timestamp --options=runtime --entitlements editor.entitlements -s ${OSX_KEY_ID} -v Godot.app/Contents/MacOS/Godot && \
               zip -r ${binname}_signed.zip Godot.app"
     
     request_uuid=$(ssh "${OSX_HOST}" "xcrun altool --notarize-app --primary-bundle-id \"${OSX_BUNDLE_ID}\" --username \"${APPLE_ID}\" --password \"${APPLE_ID_PASSWORD}\" --file ${osx_tmpdir}/${binname}_signed.zip")
--- a/build-macosx/editor.entitlements
+++ b/build-macosx/editor.entitlements
@@ -1,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<plist version="1.0">
+<dict>
+ <key>com.apple.security.device.audio-input</key>
+ <true/>
+ <key>com.apple.security.device.camera</key>
+ <true/>
+</dict>
+</plist>


build-release.sh
```diff
@@ -152,10 +152,11 @@ if [ "${build_classical}" == "1" ]; then
     osx_tmpdir=$(ssh "${OSX_HOST}" "mktemp -d")
     
     scp "${reldir}/${binname}.zip" "${OSX_HOST}:${osx_tmpdir}"
+    scp "${basedir}/build-macosx/editor.entitlements" "${OSX_HOST}:${osx_tmpdir}"
```
naithar, line 155: Not really sure if `${basedir}/build-macosx` is the correct path to use.

akien-mga:
I'm not sure about this, won't this mean that macOS will ask users to accept that Godot uses webcam and audio input regardless of their API usage? The vast majority of users don't use either feature while developing their games so it would be best if Godot was not flagged as such a privacy-intruding application.

Maybe we can instead sign without hardened runtime and let those APIs be non-functional in the editor build? As per:
https://github.com/godotengine/godot/issues/46450#issuecomment-787799816

WDYT @hpvb @bruvzg @BastiaanOlij?

BastiaanOlij:
I think that the entitlements file should be build based on a selection made by the user in the export settings. indeed if you add in that you seek entitlements for the webcam even if you don't intend to use it, the store will list webcam usage as one of the entitlements of your game. That could cost you sales. 

We already change the contents of the manifest based on whether you fill in these questions:
image

And if I recall correctly we also check whether these questions were entered when we initialize the various parts. 

So we could also build the entitlements file that way.

build-release.sh
BastiaanOlij: Should be fine as long as its accessible by code sign, the entitlements file itself does not end up in your final bundle. It's just data that is signed along with your app.

BastiaanOlij: that said I think xcode just places it in the root of your project folder.

BastiaanOlij:
owh hang on, we're signing the release binary thats in our templates? We shouldn't be signing anything until we export shouldn't we?

akien-mga:
> owh hang on, we're signing the release binary thats in our templates? We shouldn't be signing anything until we export shouldn't we?

No this is about the editor binary, not the templates.

bruvzg:
> I'm not sure about this, won't this mean that macOS will ask users to accept that Godot uses webcam and audio input regardless of their API usage? The vast majority of users don't use either feature while developing their games so it would be best if Godot was not flagged as such a privacy-intruding application.

No it shouldn't ask for permission until camera/mic is accessed for the first time.

> I think that the entitlements file should be build based on a selection made by the user in the export settings. indeed if you add in that you seek entitlements for the webcam even if you don't intend to use it, the store will list webcam usage as one of the entitlements of your game. That could cost you sales.

Would be nice to add selection to export, it's already possible for users to specify entitlements file (or disable hardened runtime), but a bunch of checkboxes is more convenient.

build-release.sh
naithar: > Should be fine as long as its accessible by code sign, the entitlements file itself does not end up in your final bundle. It's just data that is signed along with your app.

Well, it's copying `entitlements` file to remote macOS machine to make them accessible for code sign of the editor binary. The path it copies them from is the question really, since I'm not sure how script is used exactly. 
Maybe @hpvb can clarify.

BastiaanOlij:
> 
> 
> > owh hang on, we're signing the release binary thats in our templates? We shouldn't be signing anything until we export shouldn't we?
> 
> No this is about the editor binary, not the templates.

Does the editor ever call the camera code then? To be honest I think it's safest to just add the entitlements then. Its easy to explain that camera and microphone access is among the capabilities Godot offers and that just in case you use editor tools that require this access we need to add these entitlements.

bruvzg:
> Does the editor ever call the camera code then?

Only when you run the project with the camera support from the editor.

BastiaanOlij:
> No it shouldn't ask for permission until camera/mic is accessed for the first time.

It's been awhile since I last played around on iOS but I do recall the prompt indeed not coming up until you actually enable ARKit when I was building support for this.

But I thought your store entry when you upload an application to the Apple Store did display the entitlements so even if you never use the camera it would show a prospective buyer you're requesting camera access?
Or am I mixing up with the play store now?

naithar:
> I'm not sure about this, won't this mean that macOS will ask users to accept that Godot uses webcam and audio input regardless of their API usage? The vast majority of users don't use either feature while developing their games so it would be best if Godot was not flagged as such a privacy-intruding application.

As @bruvzg already mentioned the permissions are asked only when API is used. This should work the same way as unsigned editor worked, really.

> Maybe we can instead sign without hardened runtime and let those APIs be non-functional in the editor build? As per:
> [godotengine/godot#46450 (comment)](https://github.com/godotengine/godot/issues/46450#issuecomment-787799816)

I'm fine with any solution if it's working and not crashing. :) 
But I would really expect camera and microphone to work (or at least ask if I want to give it access) when running build from the editor in case I ever use this modules.

> Does the editor ever call the camera code then?

Yeah running `TestCameraServer` project from the unsigned or correctly signed  editor `.app` gives a camera permission request.

naithar:
> It's been awhile since I last played around on iOS but I do recall the prompt indeed not coming up until you actually enable ARKit when I was building support for this.
> 
> But I thought your store entry when you upload an application to the Apple Store did display the entitlements so even if you never use the camera it would show a prospective buyer you're requesting camera access?
> Or am I mixing up with the play store now?

I think you are mixing iOS and macOS permissions/entitlements :)
iOS stores permission list mostly in `.plist` while macOS seems to be requiring `entitlements` on code signing phase.

build-release.sh
akien-mga: Yeah the path seems good.

akien-mga (approved):

akien-mga:
Thanks!

PR #25, opened 2021-03-06
Make `BUILD_NAME` globally configurable

Xrayez:
Makes it easier to distinguish between official and custom builds by allowing developers to customize `BUILD_NAME` env var.

I can make it to default to `custom_build` over `official` if you like. Of course, nothing prevents someone to use `official` name, but this will decrease the likelihood of someone compiling the editor and export templates as official, otherwise the only way to distinguish such builds is by using checksums.

Also note that `BUILD_NAME` can be fetched at run-time via script with [`Engine.get_version_info()`](https://docs.godotengine.org/en/stable/classes/class_engine.html#class-engine-method-get-version-info).

### DIFF
--- a/build-android/build.sh
+++ b/build-android/build.sh
@@ -4,7 +4,6 @@ set -e
 
 # Config
 
-export BUILD_NAME=official
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
 export OPTIONS="production=yes"
 export OPTIONS_MONO="module_mono_enabled=yes mono_static=no"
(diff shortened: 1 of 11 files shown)


akien-mga:
Looks good! Yeah I think it would make sense to make it default to `custom_build`, I can set `official` in my own `config.sh`.

Xrayez:
Changed to `custom_build` by default.

akien-mga (approved):

akien-mga:
Thanks!

PR #26, opened 2021-03-06
Add `angle.7z` to `.gitignore`

Xrayez:
Note: at first I thought that it's simply not `.gitignore`d, but then I've noticed `angle` is there, but when I ran `build_release.sh`, it ended up in repository root, so makes me wonder whether it's a bug. I've prematurely made the PR instead of creating an issue about this so feel free to close this PR and fix the underlying issue, but I think merging this won't hurt anything anyways.

Here's the code:
https://github.com/godotengine/godot-build-scripts/blob/f3a5097159933becc1df63f0b6c567c2ff4830fa/build-release.sh#L263-L267

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,6 +7,7 @@ angle/
 git/
 mono-glue/
 godot.tar.gz
+angle.7z
 
 # Output
 out/


akien-mga:
It should be deleted once unzipped: https://github.com/godotengine/godot-build-scripts/blob/master/build-release.sh#L263-L267

Maybe it failed for you due to lack of `7z` and so it aborted.

Doesn't hurt yes.

Xrayez:
Ah indeed, I have `set -e` commented out so it doesn't fail in case I haven't built for all platforms (haven't managed to build for UWP anyways, see godotengine/build-containers#74).

PR #27, opened 2021-03-09
Add `Disable Library Validation` entitlement.

bruvzg:
To allow loading GDNative plug-ins, without requiring code signing.

Otherwise, loading will fail unless the library is signed by the same team ID as the app.
Screenshot 2021-03-09 at 08 06 46

With this entitlement it will allow loading, all Gatekeeper restrictions still applies to the unsigned plugins.

Fixes: https://github.com/godotengine/godot-git-plugin/issues/58

### DIFF
--- a/build-macosx/editor.entitlements
+++ b/build-macosx/editor.entitlements
@@ -6,5 +6,7 @@
  <true/>
  <key>com.apple.security.device.camera</key>
  <true/>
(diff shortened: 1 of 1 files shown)


akien-mga (approved):
Seems good, and for the reference the Mono build includes this too already (hence why it's not added here).

PR #30, opened 2021-03-29
OSX: Dummy sign templates with linker-signed

akien-mga:
This should fix potentially improper signing done by osxcross,
and allow running those on Apple M1.

---

Draft for now as I haven't actually tested this code. I also wonder if the `data_*` stuff in the Mono templates also needs to be signed with the same command or if it will be sufficient as is.

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -70,6 +70,31 @@ sign_macos() {
   fi
 }
 
+sign_macos_template() {
+  if [ -z "${OSX_HOST}" ]; then
+    return
+  fi
+  osx_tmpdir=$(ssh "${OSX_HOST}" "mktemp -d")
+  reldir="$1"
+  is_mono="$2"
+
+  if [[ "${is_mono}" == "1" ]]; then
+    extra_files="osx_template.app/Contents/Resources/data.mono.*/Mono/lib/*.dylib"
+  fi
+
+  scp "${reldir}/osx.zip" "${OSX_HOST}:${osx_tmpdir}"
+  ssh "${OSX_HOST}" "
+            cd ${osx_tmpdir} && \
+            unzip osx.zip && \
+            codesign --force -s - \
+              --options=linker-signed \
+              -v ${extra_files} osx_template.app/Contents/MacOS/* && \
+            zip -r osx_signed.zip osx_template.app"
+
+  scp "${OSX_HOST}:${osx_tmpdir}/osx_signed.zip" ${reldir}/osx.zip
+  ssh "${OSX_HOST}" "rm -rf ${osx_tmpdir}"
+}
+
 godot_version=""
 templates_version=""
 build_classical=1
@@ -209,6 +234,7 @@ if [ "${build_classical}" == "1" ]; then
   chmod +x osx_template.app/Contents/MacOS/godot_osx*
   zip -q -9 -r "${templatesdir}/osx.zip" osx_template.app
   rm -rf osx_template.app
+  sign_macos_template ${templatesdir} 0
 
   ## Server (Classical) ##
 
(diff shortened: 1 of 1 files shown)


akien-mga:
Note: The first part of the diff is mostly indentation change, review with https://github.com/godotengine/godot-build-scripts/pull/30/files?w=1 to hide whitespace changes.

build-release.sh
```diff
@@ -23,50 +23,68 @@ sign_windows() {
...
+  osx_tmpdir=$(ssh "${OSX_HOST}" "mktemp -d")
+  reldir="$1"
+
+  scp "${reldir}/osx.zip" "${OSX_HOST}:${osx_tmpdir}"
+  ssh "${OSX_HOST}" "
+            cd ${osx_tmpdir} && \
+            unzip osx.zip && \
+            codesign --force -s - \
+              --options=linker-signed \
+              -s ${OSX_KEY_ID} -v ${extra_files} ${appname} && \
```
bruvzg, line 84:
```suggestion
              -v ${extra_files} ${appname} && \
```

That's real identity, should not be here.

build-release.sh
```diff
@@ -23,50 +23,68 @@ sign_windows() {
...
+  osx_tmpdir=$(ssh "${OSX_HOST}" "mktemp -d")
+  reldir="$1"
+  binname="$2"
+  is_mono="$3"
+
+  if [[ "${is_mono}" == "1" ]]; then
+    appname="Godot_mono.app"
+    entitlements=editor_mono.entitlements
+    sharpdir="${appname}/Contents/Resources/GodotSharp"
+    extra_files="${sharpdir}/Mono/lib/*.dylib ${sharpdir}/Tools/aot-compilers/*/*"
```
bruvzg, line 38: For export templates this file list is different:

```
    sharpdir="${appname}/Contents/Resources/"
    extra_files="${sharpdir}/data.mono.osx.64.release/Mono/lib/*.dylib ${sharpdir}/data.mono.osx.64.release_debug/Mono/lib/*.dylib"
```

build-release.sh
```diff
@@ -23,50 +23,68 @@ sign_windows() {
...
+  fi
+}
+
+sign_macos_template() {
+  if [ -z "${OSX_HOST}" ]; then
+    return
   fi
+  osx_tmpdir=$(ssh "${OSX_HOST}" "mktemp -d")
+  reldir="$1"
+
```
bruvzg, line 77: `${extra_files}` and `${appname}` aren't set.

bruvzg (changes requested):

build-release.sh
akien-mga: Fixed.

build-release.sh
akien-mga: Fixed.

build-release.sh
akien-mga: That's actually the editor signing routine here. I split the cosmetic changes in a separate commit now to avoid confusion.

But if we need to sign the Mono files too for export templates, then I should re-add similar logic to `code_sign_template()`.

bruvzg: Yes, templates also need `dylib`s to be signed. This should go to the `sign_macos_template()`.

build-release.sh
```diff
@@ -23,50 +23,68 @@ sign_windows() {
...
+  scp "${reldir}/osx.zip" "${OSX_HOST}:${osx_tmpdir}"
+  ssh "${OSX_HOST}" "
+            cd ${osx_tmpdir} && \
+            unzip osx.zip && \
+            codesign --force -s - \
+              --options=linker-signed \
+              osx_template.app && \
```
bruvzg, line 84: Signing bundle won't work, its `plist` is incomplete, but `osx_template.app/Contents/Resources/data.mono.osx.64.release/Mono/lib/*.dylib osx_template.app/Contents/Resources/data.mono.osx.64.release_debug/Mono/lib/*.dylib osx_template.app/Contents/MacOS/*` will do.

It's still going to print errors about invalid `Info.plist`, but will sign all binaries, which is enough to run on arm macs.

akien-mga: Thanks, updated.

bruvzg (approved):

PR #35, opened 2021-05-05
Update build.sh

qjyqjyqjyqjy:
Give Gradlew execution permission or it will fail


akien-mga:
It already has executable permission in the Godot source code. Did you change that in your fork?

qjyqjyqjyqjy:
No more fixes, removed part does not need to build, but should not be this issue, insurance period, recommended merge.

PR #41, opened 2021-07-12
iOS: Disable non-working arm64 simulator builds

akien-mga:
Cf. https://github.com/godotengine/build-containers/pull/85

PR #42, opened 2021-07-12
Update scripts for overhauled build containers (`3.x-mono-6.12.0.147`)

akien-mga:
Cf. godotengine/build-containers#84

- Linux builds are now done on Fedora using a custom Godot SDK instead of
  relying on an old Ubuntu version for portability.
  Removes need for various workarounds.
- macOS Mono builds now include support for Apple Silicon / arm64, both
  architectures are concatenated in universal binaries.

This new config is compatible with the `3.x` branch for Godot 3.4.
It might or might not work as is for `master` / Godot 4.0. Porting it will
be the next step.

---

Also includes #41, will rebase once merged.


akien-mga:
Tested successfully on current `3.x` branch.

PR #43, opened 2021-07-13
iOS: Further changes to disable non-working arm64 simulator builds

akien-mga:
Missed in #41.

### DIFF
--- a/build-ios/build.sh
+++ b/build-ios/build.sh
@@ -45,10 +45,12 @@ if [ "${CLASSICAL}" == "1" ]; then
     IPHONESDK="/root/ioscross/x86_64_sim/SDK/iPhoneOS${IOS_SDK}.sdk" IPHONEPATH="/root/ioscross/x86_64_sim/" ios_triple="x86_64-apple-darwin11-"
 
   mkdir -p /root/out/templates
-  $IOS_LIPO -create bin/libgodot.iphone.opt.arm64.simulator.a bin/libgodot.iphone.opt.x86_64.simulator.a -output /root/out/templates/libgodot.iphone.simulator.a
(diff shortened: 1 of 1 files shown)


build-ios/build.sh
```diff
@@ -82,21 +84,26 @@ if [ "${MONO}" == "1" ]; then
...
+  #$IOS_LIPO -create bin/libmono-native.iphone.arm64.simulator.a bin/libmono-native.iphone.x86_64.simulator.a -output /root/out/templates-mono/iphone-mono-libs/libmono-native.xcframework/ios-arm64_x86_64-simulator/libmono-native.a
+  #$IOS_LIPO -create bin/libmono-profiler-log.iphone.arm64.simulator.a bin/libmono-profiler-log.iphone.x86_64.simulator.a -output /root/out/templates-mono/iphone-mono-libs/libmono-profiler-log.xcframework/ios-arm64_x86_64-simulator/libmono-profiler-log.a
+  cp bin/libmonosgen-2.0.iphone.x86_64.simulator.a /root/out/templates-mono/iphone-mono-libs/libmonosgen-2.0.xcframework/ios-x86_64-simulator/libmonosgen.a
+  cp bin/libmono-native.iphone.x86_64.simulator.a /root/out/templates-mono/iphone-mono-libs/libmono-native.xcframework/ios-x86_64-simulator/libmono-native.a
+  cp bin/libmono-profiler-log.iphone.x86_64.simulator.a /root/out/templates-mono/iphone-mono-libs/libmono-profiler-log.xcframework/ios-x86_64-simulator/libmono-profiler-log.a
```
akien-mga, line 105: @bruvzg Will `*.xcframework/ios-x86_64-simulator` work or is the `arm64_x86_64` part hardcoded in Godot?

akien-mga: Seems like it's hardcoded so I'll keep using `ios-arm64_x86_64-simulator` even though we don't have an arm64 lib.

PR #50, opened 2022-03-25
CPU architecture folder name changes

aaronfranke:
This PR changes the CPU architecture folder names for Windows and Linux. I did not touch UWP or others.

Note: I haven't actually tested this, I'm just making the assumption that this works.

Note: The reason I'm changing only the folder names first is because these changes are theoretically backportable.


akien-mga (approved):
Looks good, but I'll have to test before merging.

akien-mga:
Tested, seems to work fine :) Thanks!

PR #52, opened 2022-04-28
[Linux] Add armhf build.

Faless:
*EDIT*: Rebased, updated to use the new `arch` scons option, and removed the special `*_module_enabled=no` which are no longer necessary.

<details><summary>Original message</summary>
<p>

Some modules are disabled due to incompatibilities:
- raycast (embree), we should fix detecetion in module config.py.
- etcpak, this likely also require at least aarch64 instructions to build.
- denoise (oneDNN), also needs fix in config.py (x86_64 only).
- theora (are we keeping this?).

We should probably wait and instead of merging this PR directly, wait for https://github.com/godotengine/godot/pull/55778 or a light-weight version of it that fixes module detection (so we don't need the extra `*_module_enabled=no` shenanigan).

Depends on https://github.com/godotengine/build-containers/pull/103
Ref https://github.com/godotengine/godot-proposals/issues/988

</p>
</details>

### DIFF
--- a/build-linux/build.sh
+++ b/build-linux/build.sh
@@ -47,6 +47,19 @@ if [ "${CLASSICAL}" == "1" ]; then
   mkdir -p /root/out/x86_32/templates
   cp -rvp bin/* /root/out/x86_32/templates
   rm -rf bin
+
+  export PATH="${GODOT_SDK_LINUX_ARMHF}/bin:${BASE_PATH}"
+
+  $SCONS platform=linuxbsd arch=arm32 $OPTIONS target=editor
+  mkdir -p /root/out/arm32/tools
+  cp -rvp bin/* /root/out/arm32/tools
+  rm -rf bin
+
+  $SCONS platform=linuxbsd arch=arm32 $OPTIONS target=template_debug
+  $SCONS platform=linuxbsd arch=arm32 $OPTIONS target=template_release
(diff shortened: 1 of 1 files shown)


aaronfranke:
@Faless If you want to revisit this PR, it should be cleaner to do now that the PRs you've mentioned have been merged.

Calinou:
> denoise (oneDNN), also needs fix in config.py (x86_64 only).

Since Godot 4.2, OpenImageDenoise is no longer part of Godot itself (it's now an external program).

Faless:
> Note that we can probably build only `arm64`

The reason for doing armhf build at the time was that we didn't (and still don't) have a arm64 buildroot.

I worked on a arm64 toolchain [at the time](https://github.com/godotengine/buildroot/issues/3#issuecomment-1301979130) but it was never finalized because IIRC there were concerns about it requiring the inclusion of LLVM for vulkan/gl headers (LLVMPipe/LavaPipe). As far as I know @hpvb is almost done working on it.

I think once that's solved adding the arm64 build shouldn't be too hard.
In the meantime, I've updated this PR, but I agree that only providing arm64 builds might make sense.

akien-mga:
Superseded by #90.

For now I've opted for adding all builds (editor and templates for arm32 and arm64). We'll see if we want to reduce the list.

PR #53, opened 2022-05-10
Windows: Use osslsigncode from PATH for signing

akien-mga:
We were shipping a pre-built version but it only works on specific distros
as it's tightly coupled to openssl. When upgrading from F34 to F35 it started
segfaulting.

We now rely on the version packaged by Fedora which should be installed on the
host, or compiled and installed manually in PATH by users.

I also attempted using `osslsigncode` from the Linux container built with https://github.com/godotengine/build-containers/pull/104 but I ran into some issues:
- `build-release.sh` doesn't know about which container versions we're using (`3.x-mono-6.12.0.178` etc.), so it would need to be sourced from `build.sh`, moved to `config.sh`, or hardcoded again.
- Containers are owned by root so this would force using `sudo` for `build-release.sh` like we already do for `build.sh`, while up to now we've managed to keep `build-release.sh` as something that works for non privileged user.
- `osslsigncode` from Fedora 36 (used in containers built from https://github.com/godotengine/build-containers/pull/104) actually seems to be broken, and throws this error:

```
Failed to parse PKCS#12 file: prehensile-signing.pkcs12 (Wrong password?)
40F8BF31FF7E0000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:349:Global default library context, Algorithm (RC2-40-CBC : 0), Properties ()
Failed
```

While the same command works in a F35 container and with a locally installed F35 `osslsigncode`. That's probably because F35 uses openssl 1.1.1n while F36 uses openssl 3.0.2, and I assume osslsigncode is not compatible yet with the latter. That should be reported upstream to Fedora and osslsigncode.


akien-mga:
> * `osslsigncode` from Fedora 36 (used in containers built from https://github.com/godotengine/build-containers/pull/104     actually seems to be broken, and throws this error:
> ```
> Failed to parse PKCS#12 file: prehensile-signing.pkcs12 (Wrong password?)
> 40F8BF31FF7E0000:error:0308010C:digital envelope routines:inner_evp_generic_fetch:unsupported:crypto/evp/evp_fetch.c:349:Global default library context, Algorithm (RC2-40-CBC : 0), Properties ()
> Failed
> ```
> While the same command works in a F35 container and with a locally installed F35 osslsigncode. That's probably because F35 uses openssl 1.1.1n while F36 uses openssl 3.0.2, and I assume osslsigncode is not compatible yet with the latter. That should be reported upstream to Fedora and osslsigncode.

So it might not be a bug in the end, I was about to report it to the Fedora `osslsigncode` maintainer but I found that if I generate a PKCS#12 file using F36's openssl 3.0.2, then osslsigncode can parse it fine. So it's true that it's an incompatibility with the openssl version but it's not a bug, it just means that to use osslsigncode from F36+ we'll have to recreate our PKCS#12 file using the same openssl version.

PR #60, opened 2022-08-08
CPU architecture name changes for Windows/Linux outputs from SCons

aaronfranke:
This PR explicitly sets the `arch=` flag for Windows and Linux build commands. In the current master, explicitly setting `env["arch"]` to x86 basically only changes the file name suffix. This PR should make the build scripts forwards compatible with https://github.com/godotengine/godot/pull/55778 yet still work with the current master of Godot.

Note: I haven't tested this PR. Most of it is replacing `.64` with `.x86_64` etc so it probably works.

### DIFF
--- a/build-linux/build.sh
+++ b/build-linux/build.sh
@@ -26,26 +26,26 @@ if [ "${CLASSICAL}" == "1" ]; then
 
   export PATH="${GODOT_SDK_LINUX_X86_64}/bin:${BASE_PATH}"
 
-  $SCONS platform=linuxbsd $OPTIONS tools=yes target=release_debug
+  $SCONS platform=linuxbsd arch=x86_64 $OPTIONS tools=yes target=release_debug
   mkdir -p /root/out/x86_64/tools
   cp -rvp bin/* /root/out/x86_64/tools
   rm -rf bin
 
-  $SCONS platform=linuxbsd $OPTIONS tools=no target=release_debug
-  $SCONS platform=linuxbsd $OPTIONS tools=no target=release
+  $SCONS platform=linuxbsd arch=x86_64 $OPTIONS tools=no target=release_debug
+  $SCONS platform=linuxbsd arch=x86_64 $OPTIONS tools=no target=release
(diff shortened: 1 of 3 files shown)


build-linux/build.sh
```diff
@@ -26,26 +26,26 @@ if [ "${CLASSICAL}" == "1" ]; then
...
-  $SCONS platform=linuxbsd $OPTIONS tools=yes target=release_debug bits=32
+  $SCONS platform=linuxbsd arch=x86_32 $OPTIONS tools=yes target=release_debug bits=32
   mkdir -p /root/out/x86_32/tools
   cp -rvp bin/* /root/out/x86_32/tools
   rm -rf bin
 
-  $SCONS platform=linuxbsd $OPTIONS tools=no target=release_debug bits=32
-  $SCONS platform=linuxbsd $OPTIONS tools=no target=release bits=32
+  $SCONS platform=linuxbsd arch=x86_32 $OPTIONS tools=no target=release_debug bits=32
+  $SCONS platform=linuxbsd arch=x86_32 $OPTIONS tools=no target=release bits=32
```
akien-mga, line 48: You can also remove `bits=32`.

aaronfranke: As far as I know, `bits=32` is still necessary if we want compatibility with the current master.

akien-mga: Ah I didn't think that we currently use `arch` if it's defined over `bits` for the binary name. I thought you were trying to implement compatibility with https://github.com/godotengine/godot/pull/55778.

So yeah this might work as is.

aaronfranke: The current master has this in SConstruct:

```py
    if env["arch"] != "":
        suffix += "." + env["arch"]
    elif env["bits"] == "32":
        suffix += ".32"
    elif env["bits"] == "64":
        suffix += ".64"
```

akien-mga (approved):

akien-mga:
Thanks!

PR #62, opened 2022-08-26
JavaScript: Re-enable build for 4.0 now that WebGL 2 works

akien-mga:
I made a test build with these changes yesterday: https://downloads.tuxfamily.org/godotengine/testing/4.0.alpha15-ish.web/

I couldn't get the web editor build to render anything in Firefox or Chromium on Linux, not sure why.
It might be my CORS server deployment methods which isn't working as expected. Or maybe we regressed on WebGL 2.0 support again.

### DIFF
--- a/build-javascript/build.sh
+++ b/build-javascript/build.sh
@@ -5,8 +5,8 @@ set -e
 # Config
 
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
-export OPTIONS="production=yes"
-export OPTIONS_MONO="module_mono_enabled=yes mono_static=yes mono_prefix=/root/mono-installs/wasm-runtime-release use_lto=no"
+export OPTIONS="production=yes use_thinlto=yes"
+export OPTIONS_MONO="module_mono_enabled=yes mono_static=yes mono_prefix=/root/mono-installs/wasm-runtime-release use_lto=no use_thinlto=no"
 export TERM=xterm
 
 rm -rf godot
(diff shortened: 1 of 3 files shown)


build-javascript/build.sh
```diff
@@ -5,8 +5,8 @@ set -e
 # Config
 
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
-export OPTIONS="production=yes"
-export OPTIONS_MONO="module_mono_enabled=yes mono_static=yes mono_prefix=/root/mono-installs/wasm-runtime-release use_lto=no"
+export OPTIONS="production=yes use_thinlto=yes"
```
akien-mga, line 8: @Faless We'd have to do a proper benchmark but I think using ThinLTO for the Web platform would be better than (full) LTO. Clang/Emscripten's full LTO is single thread and just takes hours despite running on a ThreadRipper. While ThinLTO is multithreaded and takes minutes.

build-release.sh
```diff
@@ -245,23 +245,20 @@ if [ "${build_classical}" == "1" ]; then
...
+  brotli --keep --force --quality=11 ${webdir}/*
+  binname="${godot_basename}_web_editor.zip"
+  cp out/javascript/tools/godot.javascript.opt.tools.threads.zip ${reldir}/${binname}
+
+  # Templates
+  cp out/javascript/templates/godot.javascript.opt.threads.zip ${templatesdir}/webassembly_threads_release.zip
+  cp out/javascript/templates/godot.javascript.opt.debug.threads.zip ${templatesdir}/webassembly_threads_debug.zip
+
+  cp out/javascript/templates/godot.javascript.opt.gdnative.threads.zip ${templatesdir}/webassembly_gdnative_threads_release.zip
+  cp out/javascript/templates/godot.javascript.opt.debug.gdnative.threads.zip ${templatesdir}/webassembly_gdnative_threads_debug.zip
```
akien-mga, line 261: Since we now only supports threads enabled, the `_threads` suffix might be redundant? The `_gdnative` one should likely also be updated to match GDExtension. And I guess the Regular and GDNative without Threads export options should be removed.

build-javascript/build.sh
akien-mga: As seen in #66 this isn't the best option in the end so I'm reverting this change in #67.

build-release.sh
akien-mga: Solved in #67.

PR #64, opened 2022-08-29
.NET 6: Initial port of build scripts for 4.0

akien-mga:
~Draft for now, only ported Linux and Windows for testing,~ and the Linux builds are failing to link `libnethost.a` (and I suspect the Windows builds would fail like https://github.com/godotengine/godot/issues/64950).

*Edit:* Not working yet but cleans up a lot of the Mono stuff we won't need anymore. But at least all platform code is updated so I can do further refactoring without having the Mono things in the way.

Uses containers from https://github.com/godotengine/build-containers/pull/113.

### DIFF
--- a/build-android/build.sh
+++ b/build-android/build.sh
@@ -6,7 +6,7 @@ set -e
 
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
 export OPTIONS="production=yes"
-export OPTIONS_MONO="module_mono_enabled=yes mono_static=no"
+export OPTIONS_MONO="module_mono_enabled=yes"
 export TERM=xterm
 
 rm -rf godot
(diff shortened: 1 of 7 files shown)


build-android/build.sh
```diff
@@ -62,17 +62,17 @@ if [ "${MONO}" == "1" ]; then
...
 
-  $SCONS platform=android android_arch=x86 $OPTIONS $OPTIONS_MONO mono_prefix=/root/mono-installs/android-x86-release tools=no target=release_debug
-  $SCONS platform=android android_arch=x86 $OPTIONS $OPTIONS_MONO mono_prefix=/root/mono-installs/android-x86-release tools=no target=release
+  $SCONS platform=android android_arch=x86 $OPTIONS $OPTIONS_MONO tools=no target=release_debug
+  $SCONS platform=android android_arch=x86 $OPTIONS $OPTIONS_MONO tools=no target=release
 
-  $SCONS platform=android android_arch=x86_64 $OPTIONS $OPTIONS_MONO mono_prefix=/root/mono-installs/android-x86_64-release tools=no target=release_debug
-  $SCONS platform=android android_arch=x86_64 $OPTIONS $OPTIONS_MONO mono_prefix=/root/mono-installs/android-x86_64-release tools=no target=release
+  $SCONS platform=android android_arch=x86_64 $OPTIONS $OPTIONS_MONO tools=no target=release_debug
+  $SCONS platform=android android_arch=x86_64 $OPTIONS $OPTIONS_MONO tools=no target=release
```
aaronfranke, line 75: By the way, this should be `arch=x86_64`

akien-mga: Fixed in #68.

PR #67, opened 2022-09-07
Web: Handle new template names and `dlink_enabled=yes`

akien-mga:
Follow-up to https://github.com/godotengine/godot/pull/65094.

Also reverts LTO change from #62, we found that while it's much slower to link, full LTO still produces smaller binaries and that's quite important for the Web platform.


Faless (approved):
Looks good!

akien-mga:
Tested, `build-release.sh` was missing the `.wasm32` arch suffix. Now it's working fine.

PR #79, opened 2023-03-06
[3.x] Add logic to upload the Godot Android library to MavenCentral

m4gr3d:
- Address step 3 in https://github.com/godotengine/godot-proposals/issues/4230
- Add environment variables to sign the Godot Android Editor release build

**Note:** Should be merged alongside https://github.com/godotengine/godot/pull/74583

[main version](https://github.com/godotengine/godot-build-scripts/pull/78)

### DIFF
--- a/build-android/build.sh
+++ b/build-android/build.sh
@@ -14,22 +14,44 @@ mkdir godot
 cd godot
 tar xf /root/godot.tar.gz --strip-components=1
 
+# Environment variables and keystore needed for signing store editor build,
+# as well as signing and publishing to MavenCentral.
+source /root/keystore/config.sh
+
+store_release="yes"
+if [ -z "${GODOT_ANDROID_SIGN_KEYSTORE}" ]; then
+  echo "No keystore provided to sign the Android release editor build, using debug build instead."
+  store_release="no"
+fi
+
 # Classical
 
 if [ "${CLASSICAL}" == "1" ]; then
   echo "Starting classical build for Android..."
 
-  $SCONS platform=android android_arch=armv7 $OPTIONS tools=yes target=release_debug
-  $SCONS platform=android android_arch=arm64v8 $OPTIONS tools=yes target=release_debug
-  $SCONS platform=android android_arch=x86 $OPTIONS tools=yes target=release_debug
-  $SCONS platform=android android_arch=x86_64 $OPTIONS tools=yes target=release_debug
+  $SCONS platform=android android_arch=armv7 $OPTIONS tools=yes target=release_debug store_release=$store_release
+  $SCONS platform=android android_arch=arm64v8 $OPTIONS tools=yes target=release_debug store_release=$store_release
+  $SCONS platform=android android_arch=x86 $OPTIONS tools=yes target=release_debug store_release=$store_release
+  $SCONS platform=android android_arch=x86_64 $OPTIONS tools=yes target=release_debug store_release=$store_release
 
   pushd platform/android/java
   ./gradlew generateGodotEditor
   popd
 
   mkdir -p /root/out/tools
-  cp bin/android_editor.apk /root/out/tools/
+  # Copy the generated Android editor binaries (apk & aab).
+  if [ "$store_release" == "yes" ]; then
+    cp bin/android_editor_builds/android_editor-release.apk /root/out/tools/android_editor.apk
+    cp bin/android_editor_builds/android_editor-release.aab /root/out/tools/android_editor.aab
+  else
+    cp bin/android_editor_builds/android_editor-debug.apk /root/out/tools/android_editor.apk
+    cp bin/android_editor_builds/android_editor-debug.aab /root/out/tools/android_editor.aab
+  fi
+
+  # Restart from a clean tarball, as we'll copy all the contents
+  # outside the container for the MavenCentral upload.
+  rm -rf /root/godot/*
+  tar xf /root/godot.tar.gz --strip-components=1
 
   $SCONS platform=android android_arch=armv7 $OPTIONS tools=no target=release_debug
   $SCONS platform=android android_arch=armv7 $OPTIONS tools=no target=release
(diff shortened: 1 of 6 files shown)


m4gr3d:
@akien-mga This should also be included in the `3.5` branch. Should I make a separate PR for it?

akien-mga:
No need, both branches are still in sync currently.

build-android/build.sh
```diff
@@ -45,6 +46,10 @@ if [ "${CLASSICAL}" == "1" ]; then
 
   pushd platform/android/java
   ./gradlew generateGodotTemplates
+
+  # Push Godot Android library to MavenCentral
+  ./gradlew publishTemplateReleasePublicationToSonatypeRepository --max-workers 1 closeAndReleaseSonatypeStagingRepository
```
m4gr3d, line 51: Need to make it standalone and move it to the `build-release.sh` script.

build-android/build.sh
```diff
@@ -29,7 +29,8 @@ if [ "${CLASSICAL}" == "1" ]; then
   popd
 
   mkdir -p /root/out/tools
-  cp bin/android_editor.apk /root/out/tools/
+  # Note: this will fail if the apk is not signed.
+  cp bin/android_editor_builds/android_editor-release.apk /root/out/tools/
```
m4gr3d, line 33: Add a `cp` command for the aab using the same format.

build-release.sh
```diff
@@ -286,7 +286,7 @@ if [ "${build_classical}" == "1" ]; then
 
   # Editor
   binname="${godot_basename}_android_editor.apk"
-  cp out/android/tools/android_editor.apk ${reldir}/${binname}
+  cp out/android/tools/android_editor-release.apk ${reldir}/${binname}
```
m4gr3d, line 289: Add a `cp` command for the aab binary

m4gr3d:
@akien-mga Per our discussion, I've updated the MavenCentral logic to separate the upload and release steps.

build-android/build.sh
```diff
@@ -29,7 +33,15 @@ if [ "${CLASSICAL}" == "1" ]; then
   popd
 
   mkdir -p /root/out/tools
-  cp bin/android_editor.apk /root/out/tools/
+  # Copy the generated Android editor binaries (apk & aab).
+  if [ ! -e bin/android_editor_builds/android_editor-release.apk ]; then
+    echo "No signed release builds, using debug builds instead."
+    cp bin/android_editor_builds/android_editor-debug.apk /root/out/tools/android_editor.apk
+    cp bin/android_editor_builds/android_editor-debug.aab /root/out/tools/android_editor.aab
```
akien-mga, line 40: In my experience testing while I didn't have the environment variables properly sourced, when the signed release fail, what we get is actually:
```
android_editor-release-unsigned.apk
android_editor-release.aab
```
So this step fails.

akien-mga:
I updated this PR with necessary changes to make it work, after doing some testing on the build server:
- We need to source `config.sh` from `build-android/build.sh` to have the relevant values, and make the keystore visible to the build container. I opting for copying them to a dedicated folder that's exposed to the build container (so it doesn't tamper with the original files). See changes in the root `build.sh`.
- I've been struggling with some signing steps due to a special character that was being expanded in a double quotes `"` enclosed string. I changed the format to single quotes `'` to prevent this.
- Added the `.jks` keystore to `.gitignore`.

akien-mga:
Updated further, as to publish to MavenCentral in `build-release.sh`, we need to use podman to have access to the Android SDK, so we move the config of the podman binary and images version to config.sh, where they belong anyway.

build-android/build.sh
```diff
@@ -5,7 +5,7 @@ set -e
 # Config
 
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
-export OPTIONS="production=yes"
+export OPTIONS="production=yes store_release=yes"
```
m4gr3d, line 8: @akien-mga I'd updated this section and moved the `store_release` option to the actual scons command (see https://github.com/godotengine/godot-build-scripts/pull/78/files).

Having it here means we're only generating the release builds and not the debug builds.

akien-mga: That means double the build time, which is already pretty high. Do we really need those debug builds?

m4gr3d: > That means double the build time, which is already pretty high. Do we really need those debug builds?

Your call; I don't think we need them but you wanted the option of generating the debug builds for folks that are using the build scripts without a configured release key.

m4gr3d: Alternatively, either adding a descriptive comment, or checking the env variables to set the value for the `store_release` argument should do the trick without doubling the build time.

akien-mga: Yeah I'll check the env variables to set the value properly.

akien-mga:
I made a further change to fully separate the MavenCentral logic to its own script, so we no longer upload things from `build-android/build.sh`. I wasn't too happy with uploading content automatically to an official repository any time I do a test build, I prefer that this stays a manual step once things are ready for release.

So now we copy the `godot` source folder with compiled objects outside the container, so we can reuse it from `build-android/upload-mavencentral.sh` (which no longer needs passing a version string as it's not used anymore).

build-android/upload-mavencentral.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# Release the Godot Android library to MavenCentral
+${PODMAN} run -it --rm \
+  -v ${basedir}/out/android/source:/root/godot -v ${basedir}/deps/keystore:/root/keystore \
+  localhost/godot-android:${IMAGE_VERSION} bash -c \
+    "source /root/keystore/config.sh && \
+    cp -r /root/godot/.gradle /root && \
+    cd /root/godot/platform/android/java && \
+    ./gradlew publishTemplateReleasePublicationToSonatypeRepository --max-workers 1 closeSonatypeStagingRepository && \
+    ./gradlew findSonatypeStagingRepository --max-workers 1 releaseSonatypeStagingRepository"
```
m4gr3d, line 19: Since this is all being run together, both lines can be combined into:
```
./gradlew publishTemplateReleasePublicationToSonatypeRepository --max-workers 1 closeAndReleaseSonatypeStagingRepository
```

akien-mga: Thanks, fixed.

akien-mga:
Thanks!

PR #81, opened 2023-07-27
Use 7z to create more efficient ZIP archives

### LINKED ISSUE #120: Use 7-zip to create official release ZIP archives with better compression (and same compatibility)
7-zip can create fully-compatible ZIP archives that are smaller than the ones created by the reference InfoZIP tool (i.e. the `zip` command). This is done with the command `7z a -mx9 <folder>.zip <folder>`.

For instance, the official 3.2 Linux 64-bit ZIP currently weighs 28.2 MB, but it could be recompressed to weigh only 27.4 MB. The difference isn't huge, but it will likely increase as Godot's binaries keep becoming larger over time. This should also help spare TuxFamily's servers a bit, considering how often Godot is downloaded :slightly_smiling_face: 

This should help with export templates even more. It might take a while to create optimized ZIPs in this case though, as 7z is quite slow in its maximal compression mode. I still think it's worth doing for stable releases at least (and maybe prereleases if time allows).

**Edit:** I just did a test with the 3.2 Mono export templates. The ZIP created by `7z` is 480.6 MB instead of 490 MB.

**Edit 2:** On 4.0.alpha10, the 64-bit Linux editor ZIP is 46.8 MB when compressed using the above method instead of 47.9 MB. That's a saving of 1.1 MB :slightly_smiling_face:

Calinou:
For a full set of Godot 4.1.1 editor + export template builds, this saves 28 MB without affecting decompression times or archive compatibility. Any ZIP decompression software can decompress these 7z-created ZIP archives; they use the standard DEFLATE algorithm.

Creating archives is expected to be about twice as slow as `zip -9`, but given the bandwidth and download time savings, it's probably worth it.

For Mono builds, size savings are expected to be even larger as these builds are larger in the first place. Future versions will also logically benefit more from this change, as their binaries expected to keep growing over time.

**Note:** I haven't tested the script locally, so I advise performing a dry run before merging. **Both the machine `build-release.sh` is run on and the macOS machine used for signing releases must have `7z` installed and in `PATH`.** On Ubuntu, this can be done using `sudo apt install p7zip-full`. On Fedora, this can be done using `dnf install p7zip`. On macOS, this can be done using `brew install sevenzip` with [Homebrew](https://brew.sh/).

- This closes https://github.com/godotengine/godot-website/issues/120.

## Before

*Official releases as downloaded from https://godotengine.org/download/archive/4.7.1-stable/.*

```
1,280,486,955 Godot_v4.7.1-stable_export_templates.tpz
   72,992,197 Godot_v4.7.1-stable_linux.arm32.zip
   75,211,404 Godot_v4.7.1-stable_linux.arm64.zip
   76,313,228 Godot_v4.7.1-stable_linux.x86_32.zip
   76,056,717 Godot_v4.7.1-stable_linux.x86_64.zip
  166,939,858 Godot_v4.7.1-stable_macos.universal.zip
   52,243,205 Godot_v4.7.1-stable_web_editor.zip
   86,021,250 Godot_v4.7.1-stable_win32.exe.zip
   84,198,557 Godot_v4.7.1-stable_win64.exe.zip
   82,407,178 Godot_v4.7.1-stable_windows_arm64.exe.zip
```

## After

```
1,267,006,169 Godot_v4.7.1-stable_export_templates.zip
   71,765,494 Godot_v4.7.1-stable_linux.arm32.zip
   73,864,817 Godot_v4.7.1-stable_linux.arm64.zip
   75,128,599 Godot_v4.7.1-stable_linux.x86_32.zip
   74,811,081 Godot_v4.7.1-stable_linux.x86_64.zip
  164,250,615 Godot_v4.7.1-stable_macos_universal.zip
   51,334,715 Godot_v4.7.1-stable_web_editor.zip
   84,495,029 Godot_v4.7.1-stable_win32.zip
   82,660,910 Godot_v4.7.1-stable_win64.zip
   80,958,311 Godot_v4.7.1-stable_windows_arm64.zip
```

<details>
<summary>Old results (4.1.1)<summary>

## Before

*Official releases as downloaded from https://downloads.tuxfamily.org/godotengine/4.1.1/.*

```
817,195,141   Godot_v4.1.1-stable_export_templates.tpz
 47,264,118   Godot_v4.1.1-stable_linux.x86_32.zip
 54,436,186   Godot_v4.1.1-stable_linux.x86_64.zip
104,426,977   Godot_v4.1.1-stable_macos.universal.zip
 26,028,541   Godot_v4.1.1-stable_web_editor.zip
 49,422,199   Godot_v4.1.1-stable_win32.exe.zip
 54,161,086   Godot_v4.1.1-stable_win64.exe.zip
```

## After

*Official releases downloaded, extracted and recompressed with `7z a -mx9`. For the export templates TPZ, both the TPZ itself and ZIP files contained within were recompressed.*

```
797,372,026   Godot_v4.1.1-stable_export_templates.tpz
 46,153,580   Godot_v4.1.1-stable_linux.x86_32.zip
 53,220,111   Godot_v4.1.1-stable_linux.x86_64.zip
101,821,948   Godot_V4.1.1-stable_macos.universal.zip
 25,215,322   Godot_v4.1.1-stable_web_editor.zip
 48,186,220   Godot_v4.1.1-stable_win32.exe.zip
 52,936,912   Godot_v4.1.1-stable_win64.exe.zip
```
</details>

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -11,6 +11,8 @@ exec > >(tee -a "out/logs/build-release") 2>&1
 # For signing keystore and password.
 source ./config.sh
 
+export ZIP="7z a -r -tzip -mx9 -bso0 -bd"
+
 can_sign_windows=0
 if [ ! -z "${WINDOWS_SIGN_NAME}" ] && [ ! -z "${WINDOWS_SIGN_URL}" ] && [[ $(type -P "osslsigncode") ]]; then
   can_sign_windows=1
@@ -173,22 +175,22 @@ if [ "${build_classical}" == "1" ]; then
   # Editor
   binname="${godot_basename}_linux.x86_64"
   cp out/linux/x86_64/tools/godot.linuxbsd.editor.x86_64 ${binname}
-  zip -q -9 "${reldir}/${binname}.zip" ${binname}
+  $ZIP "${reldir}/${binname}.zip" ${binname}
   rm ${binname}
 
   binname="${godot_basename}_linux.x86_32"
   cp out/linux/x86_32/tools/godot.linuxbsd.editor.x86_32 ${binname}
-  zip -q -9 "${reldir}/${binname}.zip" ${binname}
+  $ZIP "${reldir}/${binname}.zip" ${binname}
   rm ${binname}
 
   binname="${godot_basename}_linux.arm64"
   cp out/linux/arm64/tools/godot.linuxbsd.editor.arm64 ${binname}
-  zip -q -9 "${reldir}/${binname}.zip" ${binname}
+  $ZIP "${reldir}/${binname}.zip" ${binname}
   rm ${binname}
 
   binname="${godot_basename}_linux.arm32"
   cp out/linux/arm32/tools/godot.linuxbsd.editor.arm32 ${binname}
-  zip -q -9 "${reldir}/${binname}.zip" ${binname}
+  $ZIP "${reldir}/${binname}.zip" ${binname}
   rm ${binname}
 
   # ICU data
@@ -217,7 +219,7 @@ if [ "${build_classical}" == "1" ]; then
   sign_windows ${binname}
   cp out/windows/x86_64/tools/godot.windows.editor.x86_64.console.exe ${wrpname}
   sign_windows ${wrpname}
-  zip -q -9 "${reldir}/${binname}.zip" ${binname} ${wrpname}
+  $ZIP "${reldir}/${binname}.zip" ${binname} ${wrpname}
   rm ${binname} ${wrpname}
 
   binname="${godot_basename}_win32.exe"
@@ -226,7 +228,7 @@ if [ "${build_classical}" == "1" ]; then
   sign_windows ${binname}
   cp out/windows/x86_32/tools/godot.windows.editor.x86_32.console.exe ${wrpname}
   sign_windows ${wrpname}
-  zip -q -9 "${reldir}/${binname}.zip" ${binname} ${wrpname}
+  $ZIP "${reldir}/${binname}.zip" ${binname} ${wrpname}
   rm ${binname} ${wrpname}
 
   binname="${godot_basename}_windows_arm64.exe"
@@ -235,7 +237,7 @@ if [ "${build_classical}" == "1" ]; then
   sign_windows ${binname}
   cp out/windows/arm64/tools/godot.windows.editor.arm64.llvm.console.exe ${wrpname}
   sign_windows ${wrpname}
-  zip -q -9 "${reldir}/${binname}.zip" ${binname} ${wrpname}
+  $ZIP "${reldir}/${binname}.zip" ${binname} ${wrpname}
   rm ${binname} ${wrpname}
 
   # Templates
@@ -262,7 +264,7 @@ if [ "${build_classical}" == "1" ]; then
   cp out/macos/tools/godot.macos.editor.universal Godot.app/Contents/MacOS/Godot
   chmod +x Godot.app/Contents/MacOS/Godot
   sign_macos Godot.app
-  zip -q -9 -r "${reldir}/${binname}.zip" Godot.app
+  $ZIP "${reldir}/${binname}.zip" Godot.app
   rm -rf Godot.app
 
   # Templates
@@ -274,7 +276,7 @@ if [ "${build_classical}" == "1" ]; then
   cp out/macos/templates/godot.macos.template_debug.universal macos_template.app/Contents/MacOS/godot_macos_debug.universal
   chmod +x macos_template.app/Contents/MacOS/godot_macos*
   sign_macos_template macos_template.app
-  zip -q -9 -r "${templatesdir}/macos.zip" macos_template.app
+  $ZIP "${templatesdir}/macos.zip" macos_template.app
   rm -rf macos_template.app
 
   ## Steam (Classical) ##
@@ -302,7 +304,7 @@ if [ "${build_classical}" == "1" ]; then
     cp deps/steam/libsteam_api.dylib Godot.app/Contents/Frameworks/libsteam_api.dylib
     chmod +x Godot.app/Contents/MacOS/Godot
     sign_macos Godot.app
-    zip -q -9 -r "${binname}_steam.zip" Godot.app
+    $ZIP "${binname}_steam.zip" Godot.app
     rm -rf Godot.app
     unzip ${binname}_steam.zip -d ${steamdir}/
     rm -f ${binname}_steam.zip
@@ -372,7 +374,7 @@ if [ "${build_classical}" == "1" ]; then
   cp -r deps/moltenvk/MoltenVK/MoltenVK.xcframework ios_xcode/
   rm -rf ios_xcode/MoltenVK.xcframework/{macos,tvos,xros}*
   cd ios_xcode
-  zip -q -9 -r "${templatesdir}/ios.zip" *
+  $ZIP "${templatesdir}/ios.zip" *
   cd ..
   rm -rf ios_xcode
 
@@ -394,7 +396,7 @@ if [ "${build_classical}" == "1" ]; then
 
   echo "${templates_version}" > ${templatesdir}/version.txt
   pushd ${templatesdir}/..
-  zip -q -9 -r -D "${reldir}/${godot_basename}_export_templates.tpz" templates/*
+  $ZIP "${reldir}/${godot_basename}_export_templates.tpz" templates/*
   popd
 
   ## SHA-512 sums (Classical) ##
@@ -418,28 +420,28 @@ if [ "${build_mono}" == "1" ]; then
   mkdir -p ${binbasename}_x86_64
   cp out/linux/x86_64/tools-mono/godot.linuxbsd.editor.x86_64.mono ${binbasename}_x86_64/${binbasename}.x86_64
   cp -rp out/linux/x86_64/tools-mono/GodotSharp ${binbasename}_x86_64/
-  zip -r -q -9 "${reldir_mono}/${binbasename}_x86_64.zip" ${binbasename}_x86_64
+  $ZIP "${reldir_mono}/${binbasename}_x86_64.zip" ${binbasename}_x86_64
   rm -rf ${binbasename}_x86_64
 
   binbasename="${godot_basename}_mono_linux"
   mkdir -p ${binbasename}_x86_32
   cp out/linux/x86_32/tools-mono/godot.linuxbsd.editor.x86_32.mono ${binbasename}_x86_32/${binbasename}.x86_32
   cp -rp out/linux/x86_32/tools-mono/GodotSharp/ ${binbasename}_x86_32/
-  zip -r -q -9 "${reldir_mono}/${binbasename}_x86_32.zip" ${binbasename}_x86_32
+  $ZIP "${reldir_mono}/${binbasename}_x86_32.zip" ${binbasename}_x86_32
   rm -rf ${binbasename}_x86_32
 
   binbasename="${godot_basename}_mono_linux"
   mkdir -p ${binbasename}_arm64
   cp out/linux/arm64/tools-mono/godot.linuxbsd.editor.arm64.mono ${binbasename}_arm64/${binbasename}.arm64
   cp -rp out/linux/arm64/tools-mono/GodotSharp/ ${binbasename}_arm64/
-  zip -r -q -9 "${reldir_mono}/${binbasename}_arm64.zip" ${binbasename}_arm64
+  $ZIP "${reldir_mono}/${binbasename}_arm64.zip" ${binbasename}_arm64
   rm -rf ${binbasename}_arm64
 
   binbasename="${godot_basename}_mono_linux"
   mkdir -p ${binbasename}_arm32
   cp out/linux/arm32/tools-mono/godot.linuxbsd.editor.arm32.mono ${binbasename}_arm32/${binbasename}.arm32
   cp -rp out/linux/arm32/tools-mono/GodotSharp/ ${binbasename}_arm32/
-  zip -r -q -9 "${reldir_mono}/${binbasename}_arm32.zip" ${binbasename}_arm32
+  $ZIP "${reldir_mono}/${binbasename}_arm32.zip" ${binbasename}_arm32
   rm -rf ${binbasename}_arm32
 
   # ICU data
@@ -470,7 +472,7 @@ if [ "${build_mono}" == "1" ]; then
   cp -rp out/windows/x86_64/tools-mono/GodotSharp ${binname}/
   cp out/windows/x86_64/tools-mono/godot.windows.editor.x86_64.mono.console.exe ${binname}/${wrpname}.exe
   sign_windows ${binname}/${wrpname}.exe
-  zip -r -q -9 "${reldir_mono}/${binname}.zip" ${binname}
+  $ZIP "${reldir_mono}/${binname}.zip" ${binname}
   rm -rf ${binname}
 
   binname="${godot_basename}_mono_win32"
@@ -481,7 +483,7 @@ if [ "${build_mono}" == "1" ]; then
   cp -rp out/windows/x86_32/tools-mono/GodotSharp ${binname}/
   cp out/windows/x86_32/tools-mono/godot.windows.editor.x86_32.mono.console.exe ${binname}/${wrpname}.exe
   sign_windows ${binname}/${wrpname}.exe
-  zip -r -q -9 "${reldir_mono}/${binname}.zip" ${binname}
+  $ZIP "${reldir_mono}/${binname}.zip" ${binname}
   rm -rf ${binname}
 
   binname="${godot_basename}_mono_windows_arm64"
@@ -492,7 +494,7 @@ if [ "${build_mono}" == "1" ]; then
   cp -rp out/windows/arm64/tools-mono/GodotSharp ${binname}/
   cp out/windows/arm64/tools-mono/godot.windows.editor.arm64.llvm.mono.console.exe ${binname}/${wrpname}.exe
   sign_windows ${binname}/${wrpname}.exe
-  zip -r -q -9 "${reldir_mono}/${binname}.zip" ${binname}
+  $ZIP "${reldir_mono}/${binname}.zip" ${binname}
   rm -rf ${binname}
 
   # Templates
@@ -520,7 +522,7 @@ if [ "${build_mono}" == "1" ]; then
   cp -rp out/macos/tools-mono/GodotSharp Godot_mono.app/Contents/Resources/GodotSharp
   chmod +x Godot_mono.app/Contents/MacOS/Godot
   sign_macos Godot_mono.app
-  zip -q -9 -r "${reldir_mono}/${binname}.zip" Godot_mono.app
+  $ZIP "${reldir_mono}/${binname}.zip" Godot_mono.app
   rm -rf Godot_mono.app
 
   # Templates
@@ -531,7 +533,7 @@ if [ "${build_mono}" == "1" ]; then
   cp out/macos/templates-mono/godot.macos.template_release.universal.mono macos_template.app/Contents/MacOS/godot_macos_release.universal
   chmod +x macos_template.app/Contents/MacOS/godot_macos*
   sign_macos_template macos_template.app
-  zip -q -9 -r "${templatesdir_mono}/macos.zip" macos_template.app
+  $ZIP "${templatesdir_mono}/macos.zip" macos_template.app
   rm -rf macos_template.app
 
   ## Android (Mono) ##
@@ -554,11 +556,11 @@ if [ "${build_mono}" == "1" ]; then
   cp out/ios/templates-mono/libgodot.ios.a ios_xcode/libgodot.ios.release.xcframework/ios-arm64/libgodot.a
   cp out/ios/templates-mono/libgodot.ios.debug.a ios_xcode/libgodot.ios.debug.xcframework/ios-arm64/libgodot.a
   cp out/ios/templates-mono/libgodot_camera.ios.a ios_xcode/libgodot_camera.ios.release.xcframework/ios-arm64/libgodot_camera.a
-  cp out/ios/templates-mono/libgodot_camera.ios.debug.a ios_xcode/libgodot_camera.ios.debug.xcframework/ios-arm64/libgodot_camera.a 
+  cp out/ios/templates-mono/libgodot_camera.ios.debug.a ios_xcode/libgodot_camera.ios.debug.xcframework/ios-arm64/libgodot_camera.a
   cp -r deps/moltenvk/MoltenVK/MoltenVK.xcframework ios_xcode/
   rm -rf ios_xcode/MoltenVK.xcframework/{macos,tvos}*
   cd ios_xcode
-  zip -q -9 -r "${templatesdir_mono}/ios.zip" *
+  $ZIP "${templatesdir_mono}/ios.zip" *
   cd ..
   rm -rf ios_xcode
 
@@ -591,7 +593,7 @@ if [ "${build_mono}" == "1" ]; then
 
   echo "${templates_version}.mono" > ${templatesdir_mono}/version.txt
   pushd ${templatesdir_mono}/..
-  zip -q -9 -r -D "${reldir_mono}/${godot_basename}_mono_export_templates.tpz" templates/*
+  $ZIP "${reldir_mono}/${godot_basename}_mono_export_templates.tpz" templates/*
   popd
 
   ## SHA-512 sums (Mono) ##
@@ -614,7 +616,7 @@ if [ "${build_dotnet}" == "1" ]; then
     # Editor
     binname="${godot_basename}_dotnet_linux.${arch}"
     cp out/linux/${arch}/tools-dotnet/godot.linuxbsd.editor.${arch}.dotnet ${binname}
-    zip -r -q -9 "${reldir_dotnet}/${binname}.zip" ${binname}
+    $ZIP "${reldir_dotnet}/${binname}.zip" ${binname}
     rm ${binname}
 
     # Templates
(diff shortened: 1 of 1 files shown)


build-release.sh
```diff
@@ -219,13 +219,13 @@ if [ "${build_classical}" == "1" ]; then
   binname="${godot_basename}_linux.x86_64"
   cp out/linux/x86_64/tools/godot.linuxbsd.editor.x86_64 ${binname}
   strip ${binname}
-  zip -q -9 "${reldir}/${binname}.zip" ${binname}
+  7z a -bso0 -bd -mx9 "${reldir}/${binname}.zip" ${binname}
```
akien-mga, line 222: Can you make this a define like:
```
export ZIP="7z a -bso0 -bd -mx9"
```

to use with `${ZIP}` and `${ZIP} -r`, so we don't have to hardcode the options everywhere?

build-release.sh
```diff
@@ -47,7 +49,7 @@ sign_macos() {
             codesign --force --timestamp \
               --options=runtime --entitlements editor.entitlements \
               -s ${OSX_KEY_ID} -v ${_appname} && \
-            zip -r ${_binname}_signed.zip ${_appname}"
+            $ZIP ${_binname}_signed.zip ${_appname}"
```
akien-mga, line 52: Missing `-r`, no? Same for some others below.

build-release.sh
```diff
@@ -83,7 +85,7 @@ sign_macos_template() {
             codesign --force -s - \
               --options=linker-signed \
               -v macos_template.app/Contents/MacOS/* && \
-            zip -r macos_signed.zip macos_template.app"
+            $ZIP macos_signed.zip macos_template.app"
```
akien-mga, line 88: Missing `-r`

build-release.sh
```diff
@@ -360,15 +362,15 @@ if [ "${build_classical}" == "1" ]; then
...
+  $ZIP -r "${templatesdir}/ios.zip" *
   cd ..
   rm -rf ios_xcode
 
   ## Templates TPZ (Classical) ##
 
   echo "${templates_version}" > ${templatesdir}/version.txt
   pushd ${templatesdir}/..
-  zip -q -9 -r -D "${reldir}/${godot_basename}_export_templates.tpz" templates/*
+  $ZIP -r "${reldir}/${godot_basename}_export_templates.tpz" templates/*
```
akien-mga, line 373: I think `-D` was important here, so this might break the logic.

build-release.sh
```diff
@@ -437,7 +439,7 @@ if [ "${build_mono}" == "1" ]; then
   cp out/windows/x86_32/tools-mono/godot.windows.editor.x86_32.mono.console.exe ${binname}/${wrpname}.exe
   strip ${binname}/${wrpname}.exe
   sign_windows ${binname}/${wrpname}.exe
-  zip -r -q -9 "${reldir_mono}/${binname}.zip" ${binname}
+  $ZIP "${reldir_mono}/${binname}.zip" ${binname}
```
akien-mga, line 442: -4

build-release.sh
```diff
@@ -516,7 +518,7 @@ if [ "${build_mono}" == "1" ]; then
 
   echo "${templates_version}.mono" > ${templatesdir_mono}/version.txt
   pushd ${templatesdir_mono}/..
-  zip -q -9 -r -D "${reldir_mono}/${godot_basename}_mono_export_templates.tpz" templates/*
+  $ZIP -r "${reldir_mono}/${godot_basename}_mono_export_templates.tpz" templates/*
```
akien-mga, line 521: Also needs -D equivalent.

build-release.sh
Calinou: `7z` doesn't seem to have an equivalent of `-D`. According to `zip`'s command line help:

```
 -D   do not add directory entries
```

The only difference I can notice between 2 ZIP archives created (one with `-D`, one without `-D`) is that the former will have directory entries in its `7z l` listing, but the latter will specify full paths to every file.

In practice, this appears to make no concrete difference on what you see when opening the ZIP archive: both have the exact same folder structure when viewed with GUI archive managers. In short, `-D` does **not** make the ZIP's structure flat or remove the top-level folder as the option's name might suggest.

I think the `-D` requirement is a leftover from early Godot 3.x days where Godot behaved poorly if given a TPZ that had directory entries.

akien-mga:
Needs rebase after #90. Then I'll test and see if I can merge for the next beta.

build-release.sh
```diff
@@ -47,7 +49,7 @@ sign_macos() {
             codesign --force --timestamp \
               --options=runtime --entitlements editor.entitlements \
               -s ${OSX_KEY_ID} -v ${_appname} && \
-            zip -r ${_binname}_signed.zip ${_appname}"
+            $ZIP -r ${_binname}_signed.zip ${_appname}"
```
akien-mga, line 52: So this bit forces me to have 7zip installed on macOS, which is less trivial than on Linux.

It seems available on homebrew but homebrew isn't set up on HP's Mac, could likely install it.

But this begs the question of whether there's anything more actively maintained than 7zip to compress zips. The last release dates back to 2016.

Calinou: > But this begs the question of whether there's anything more actively maintained than 7zip to compress zips. The last release dates back to 2016.

The last GUI release was made in June 2023: https://7-zip.org/
However, the CLI version was never updated past 16.02 (May 2016). This is an unfortunate artifact of 7-zip being developed around its Windows GUI first, and also being developed behind closed doors.

I don't know of any alternatives that have the same level of efficiency and performance for compressing ZIP archives. Using Zopfli to create a ZIP archive will result in slightly more efficiency, but it's more than 10× slower which makes it not worth the time.

akien-mga: > The last GUI release was made in June 2023: [7-zip.org](https://7-zip.org/)

It's Windows only though, no?

I found PeaZip which seems actively maintained and properly cross-platform: https://github.com/peazip/PeaZip/releases/

Might be worth comparing with 7zip.

bruvzg: > But this begs the question of whether there's anything more actively maintained than 7zip to compress zips. The last release dates back to 2016.

At least of macOS homebrew, there are two 7Z packages:
- `p7zip` - `7z` executable, built from p7zip port source from 2017 (17.04).
- `sevenzip` - `7zz` executable, built from official 2023 source (23.01, same as GUI version).

Seems like official source added support for Linux/macOS command line build in 2021.

DeanLemans:
Why is this pr not yet merged?

DeanLemans (approved):

Calinou:
Rebased. I'll run 4.7.1 binary size comparison tests shortly (since the binaries are larger, the size benefit from this PR will also be larger).

**Edit:** Done (see OP).

akien-mga:
CC @Repiteo 

Sorry for the delay reviewing this. I've been worried from the start about how much slower the compression would be with 7z, given that it's already pretty slow with zip and we're trying to compress more.

I've now done some tests on debug symbols which will make the problem an order of magnitude worse: https://github.com/godotengine/godot-build-scripts/pull/103#issuecomment-5340697808

And it seems like 7zip DEFLATE compression with `-mx9` is around 5 times slower than `zip -9`, at least on those Linux debug symbols. For a gain that's quite minimal overall.

We can take more time to compress stuff on the build server but we should still make sure the time to package a release don't balloon to several hours.

I think there's still merit in using better compression here but we should do a size/compression time comparison to pick the best tradeoff. (Also, from a quick try 7zip doesn't seem to be able to use multiple threads despite its `-mmt=N` argument, I assume this only works for 7z format and not .zip?)

I think what we might want to evaluate to be really efficient:
- Compress the templates .tpz with a better format like zstd, which we support in the engine. But I don't know if the method we use to extract specific templates from the .tpz with the new Export Templates Manager can work with non-zip formats.
- Compress the editor .zips with 7z and a compression ratio that has a good tradeoff. We could go for `-mx9` here if it's only for editor builds, but if there's not much gain compared to lower factors and a significant increase in compression time, we can also pick something more reasonable.
  * We keep .zip here for convenience so it's easy to extract for end users (unlike .7z, .tar.xz or .tar.zst, etc.). Though it might be worth reviewing whether Windows 10/11 now support better formats out of the box, in which case we could consider using something better.
- Compress the debug symbols with zstd.

build-release.sh
```diff
@@ -302,7 +304,7 @@ if [ "${build_classical}" == "1" ]; then
     cp deps/steam/libsteam_api.dylib Godot.app/Contents/Frameworks/libsteam_api.dylib
     chmod +x Godot.app/Contents/MacOS/Godot
     sign_macos Godot.app
-    zip -q -9 -r "${binname}_steam.zip" Godot.app
+    $ZIP "${binname}_steam.zip" Godot.app
```
akien-mga, line 307:
```suggestion
    $ZIP -r "${binname}_steam.zip" Godot.app
```

build-release.sh
```diff
@@ -418,28 +420,28 @@ if [ "${build_mono}" == "1" ]; then
...
-  zip -r -q -9 "${reldir_mono}/${binbasename}_x86_32.zip" ${binbasename}_x86_32
+  $ZIP -r "${reldir_mono}/${binbasename}_x86_32.zip" ${binbasename}_x86_32
   rm -rf ${binbasename}_x86_32
 
   binbasename="${godot_basename}_mono_linux"
   mkdir -p ${binbasename}_arm64
   cp out/linux/arm64/tools-mono/godot.linuxbsd.editor.arm64.mono ${binbasename}_arm64/${binbasename}.arm64
   cp -rp out/linux/arm64/tools-mono/GodotSharp/ ${binbasename}_arm64/
-  zip -r -q -9 "${reldir_mono}/${binbasename}_arm64.zip" ${binbasename}_arm64
+  $ZIP "${reldir_mono}/${binbasename}_arm64.zip" ${binbasename}_arm64
```
akien-mga, line 437:
```suggestion
  $ZIP -r "${reldir_mono}/${binbasename}_arm64.zip" ${binbasename}_arm64
```

build-release.sh
```diff
@@ -418,28 +420,28 @@ if [ "${build_mono}" == "1" ]; then
...
   rm -rf ${binbasename}_arm64
 
   binbasename="${godot_basename}_mono_linux"
   mkdir -p ${binbasename}_arm32
   cp out/linux/arm32/tools-mono/godot.linuxbsd.editor.arm32.mono ${binbasename}_arm32/${binbasename}.arm32
   cp -rp out/linux/arm32/tools-mono/GodotSharp/ ${binbasename}_arm32/
-  zip -r -q -9 "${reldir_mono}/${binbasename}_arm32.zip" ${binbasename}_arm32
+  $ZIP "${reldir_mono}/${binbasename}_arm32.zip" ${binbasename}_arm32
```
akien-mga, line 444:
```suggestion
  $ZIP -r "${reldir_mono}/${binbasename}_arm32.zip" ${binbasename}_arm32
```

build-release.sh
akien-mga: Not fixed.

build-release.sh
```diff
@@ -614,7 +616,7 @@ if [ "${build_dotnet}" == "1" ]; then
     # Editor
     binname="${godot_basename}_dotnet_linux.${arch}"
     cp out/linux/${arch}/tools-dotnet/godot.linuxbsd.editor.${arch}.dotnet ${binname}
-    zip -r -q -9 "${reldir_dotnet}/${binname}.zip" ${binname}
+    $ZIP "${reldir_dotnet}/${binname}.zip" ${binname}
```
akien-mga, line 619:
```suggestion
    $ZIP -r "${reldir_dotnet}/${binname}.zip" ${binname}
```

build-release.sh
```diff
@@ -647,7 +649,7 @@ if [ "${build_dotnet}" == "1" ]; then
     sign_windows ${binname}
     cp out/windows/${arch}/tools-dotnet/godot.windows.editor.${arch}${is_llvm}.dotnet.console.exe ${wrpname}
     sign_windows ${wrpname}
-    zip -r -q -9 "${reldir_dotnet}/${binname}.zip" ${binname} ${wrpname}
+    $ZIP "${reldir_dotnet}/${binname}.zip" ${binname} ${wrpname}
```
akien-mga, line 652:
```suggestion
    $ZIP -r "${reldir_dotnet}/${binname}.zip" ${binname} ${wrpname}
```

build-release.sh
```diff
@@ -11,6 +11,8 @@ exec > >(tee -a "out/logs/build-release") 2>&1
 # For signing keystore and password.
 source ./config.sh
 
+export ZIP="7z a -bso0 -bd -mx9"
```
akien-mga: We could likely include `-r` here and have it always included, I suppose it's a no-op for files.
```suggestion
export ZIP="7z a -r -bso0 -bd -mx9"
```

build-release.sh
akien-mga: Took me 3 years to get back to testing this, but it seems I was right 3 years ago...

This line fails with:
```
System ERROR:
E_INVALIDARG
```

akien-mga: Though the problem is not what I thought, but the fact that 7zip can't guess we want a zip/DEFLATE compression with a `.tpz` extension. We need to specify `-tzip` explicitly.

bruvzg: It's likely failing because of `tpz` in the output file name (unknown format), not missing `-D`.

build-release.sh
bruvzg: `7z` (2017 version) vs `7zz` (current version) seems still be the case. But compressing 4.8dev3 templates, both produce identical output.

akien-mga: We don't need to run 7z on macOS anymore either way, now rcodesign gives us back the `.app` after signing and notarization and we zip it on Linux.

But interesting to learn that there's now a proper upstream `7zz` port for Linux and macOS, we should probably use that instead of the old `p7zip` if we're going to use 7zip.

akien-mga:
Tested `build-release.sh` for 4.7.2-stable with code signing/notarization disabled.

|Compression|Time|Size|
|---|---|---|
|`zip -9`|11m 39s|7.935 GiB|
|`7z -tzip -mx9`|31m 26s|7.853 GiB|

So we're saving around 1% of disk size for 3 times the packaging time.
IMO it's not worth it as is.

<details>
<summary>Build sizes for all files</summary>

4.7.2-stable with `zip -9`:
```
total 6092588
-rw-r--r--. 1 akien akien   47744300 Aug 19 12:48 godot-4.7.2-stable.tar.xz
-rw-r--r--. 1 akien akien         92 Aug 19 12:48 godot-4.7.2-stable.tar.xz.sha256
-rw-r--r--. 1 akien akien  103251267 Aug 19 12:48 godot-lib.4.7.2.stable.template_release.aar
-rw-r--r--. 1 akien akien  105863162 Aug 19 12:48 godot-lib.4.7.2.stable.template_release.perfetto.aar
-rw-r--r--. 1 akien akien 1309969445 Aug 19 12:48 Godot_native_debug_symbols.4.7.2.stable.editor.android.zip
-rw-r--r--. 1 akien akien  780497672 Aug 19 12:48 Godot_native_debug_symbols.4.7.2.stable.template_release.android.zip
-rw-r--r--. 1 akien akien  130087617 Aug 19 12:48 Godot_v4.7.2-stable_android_debug.perfetto.apk
-rw-r--r--. 1 akien akien  331935184 Aug 19 12:48 Godot_v4.7.2-stable_android_editor.aab
-rw-r--r--. 1 akien akien  688166381 Aug 19 12:48 Godot_v4.7.2-stable_android_editor.apk
-rw-r--r--. 1 akien akien  168981853 Aug 19 12:48 Godot_v4.7.2-stable_android_editor_horizonos.apk
-rw-r--r--. 1 akien akien  176377894 Aug 19 12:48 Godot_v4.7.2-stable_android_editor_picoos.apk
-rw-r--r--. 1 akien akien  107415229 Aug 19 12:48 Godot_v4.7.2-stable_android_release.perfetto.apk
-rw-r--r--. 1 akien akien  219513141 Aug 19 12:48 Godot_v4.7.2-stable_android_source.perfetto.zip
-rw-r--r--. 1 akien akien 1279884195 Aug 19 12:48 Godot_v4.7.2-stable_export_templates.tpz
-rw-r--r--. 1 akien akien   74780990 Aug 19 12:48 Godot_v4.7.2-stable_linux.arm32.zip
-rw-r--r--. 1 akien akien   77008699 Aug 19 12:48 Godot_v4.7.2-stable_linux.arm64.zip
-rw-r--r--. 1 akien akien   78111204 Aug 19 12:48 Godot_v4.7.2-stable_linux.x86_32.zip
-rw-r--r--. 1 akien akien   77860424 Aug 19 12:48 Godot_v4.7.2-stable_linux.x86_64.zip
-rw-r--r--. 1 akien akien  169201218 Aug 19 12:48 Godot_v4.7.2-stable_macos.universal.zip
-rw-r--r--. 1 akien akien   54031095 Aug 19 12:48 Godot_v4.7.2-stable_web_editor.zip
-rw-r--r--. 1 akien akien   87823657 Aug 19 12:48 Godot_v4.7.2-stable_win32.exe.zip
-rw-r--r--. 1 akien akien   85999453 Aug 19 12:48 Godot_v4.7.2-stable_win64.exe.zip
-rw-r--r--. 1 akien akien   84249172 Aug 19 12:48 Godot_v4.7.2-stable_windows_arm64.exe.zip
-rw-r--r--. 1 akien akien       3963 Aug 19 12:48 SHA512-SUMS.txt

mono:
total 2227560
-rw-r--r--. 1 akien akien  103605746 Aug 19 12:48 godot-lib.4.7.2.stable.mono.template_release.aar
-rw-r--r--. 1 akien akien 1201176359 Aug 19 12:48 Godot_v4.7.2-stable_mono_export_templates.tpz
-rw-r--r--. 1 akien akien  104613285 Aug 19 12:48 Godot_v4.7.2-stable_mono_linux_arm32.zip
-rw-r--r--. 1 akien akien  106869118 Aug 19 12:48 Godot_v4.7.2-stable_mono_linux_arm64.zip
-rw-r--r--. 1 akien akien  107861168 Aug 19 12:48 Godot_v4.7.2-stable_mono_linux_x86_32.zip
-rw-r--r--. 1 akien akien  107698034 Aug 19 12:48 Godot_v4.7.2-stable_mono_linux_x86_64.zip
-rw-r--r--. 1 akien akien  199323560 Aug 19 12:48 Godot_v4.7.2-stable_mono_macos.universal.zip
-rw-r--r--. 1 akien akien  118415787 Aug 19 12:48 Godot_v4.7.2-stable_mono_win32.zip
-rw-r--r--. 1 akien akien  116567578 Aug 19 12:48 Godot_v4.7.2-stable_mono_win64.zip
-rw-r--r--. 1 akien akien  114866261 Aug 19 12:48 Godot_v4.7.2-stable_mono_windows_arm64.zip
-rw-r--r--. 1 akien akien       1719 Aug 19 12:48 SHA512-SUMS.txt
```

4.7.2-stable with `7z -tzip -mx9`:
```

total 6052728
-rw-r--r--. 1 akien akien   47744300 Aug 19 13:41 godot-4.7.2-stable.tar.xz
-rw-r--r--. 1 akien akien         92 Aug 19 13:41 godot-4.7.2-stable.tar.xz.sha256
-rw-r--r--. 1 akien akien  103251267 Aug 19 13:56 godot-lib.4.7.2.stable.template_release.aar
-rw-r--r--. 1 akien akien  105863162 Aug 19 13:56 godot-lib.4.7.2.stable.template_release.perfetto.aar
-rw-r--r--. 1 akien akien 1309969445 Aug 19 13:56 Godot_native_debug_symbols.4.7.2.stable.editor.android.zip
-rw-r--r--. 1 akien akien  780497672 Aug 19 13:56 Godot_native_debug_symbols.4.7.2.stable.template_release.android.zip
-rw-r--r--. 1 akien akien  130087617 Aug 19 13:56 Godot_v4.7.2-stable_android_debug.perfetto.apk
-rw-r--r--. 1 akien akien  331935184 Aug 19 13:56 Godot_v4.7.2-stable_android_editor.aab
-rw-r--r--. 1 akien akien  688166381 Aug 19 13:56 Godot_v4.7.2-stable_android_editor.apk
-rw-r--r--. 1 akien akien  168981853 Aug 19 13:56 Godot_v4.7.2-stable_android_editor_horizonos.apk
-rw-r--r--. 1 akien akien  176377894 Aug 19 13:56 Godot_v4.7.2-stable_android_editor_picoos.apk
-rw-r--r--. 1 akien akien  107415229 Aug 19 13:56 Godot_v4.7.2-stable_android_release.perfetto.apk
-rw-r--r--. 1 akien akien  219513141 Aug 19 13:56 Godot_v4.7.2-stable_android_source.perfetto.zip
-rw-r--r--. 1 akien akien 1251287721 Aug 19 13:59 Godot_v4.7.2-stable_export_templates.tpz
-rw-r--r--. 1 akien akien   73554543 Aug 19 13:44 Godot_v4.7.2-stable_linux.arm32.zip
-rw-r--r--. 1 akien akien   75664530 Aug 19 13:43 Godot_v4.7.2-stable_linux.arm64.zip
-rw-r--r--. 1 akien akien   76927489 Aug 19 13:43 Godot_v4.7.2-stable_linux.x86_32.zip
-rw-r--r--. 1 akien akien   76617119 Aug 19 13:42 Godot_v4.7.2-stable_linux.x86_64.zip
-rw-r--r--. 1 akien akien  166496467 Aug 19 13:50 Godot_v4.7.2-stable_macos.universal.zip
-rw-r--r--. 1 akien akien   54031095 Aug 19 13:56 Godot_v4.7.2-stable_web_editor.zip
-rw-r--r--. 1 akien akien   86291978 Aug 19 13:47 Godot_v4.7.2-stable_win32.exe.zip
-rw-r--r--. 1 akien akien   84461826 Aug 19 13:45 Godot_v4.7.2-stable_win64.exe.zip
-rw-r--r--. 1 akien akien   82793483 Aug 19 13:48 Godot_v4.7.2-stable_windows_arm64.exe.zip
-rw-r--r--. 1 akien akien       3963 Aug 19 13:59 SHA512-SUMS.txt

mono:
total 2182024
-rw-r--r--. 1 akien akien  103605746 Aug 19 14:09 godot-lib.4.7.2.stable.mono.template_release.aar
-rw-r--r--. 1 akien akien 1172173746 Aug 19 14:12 Godot_v4.7.2-stable_mono_export_templates.tpz
-rw-r--r--. 1 akien akien  102722984 Aug 19 14:02 Godot_v4.7.2-stable_mono_linux_arm32.zip
-rw-r--r--. 1 akien akien  104868467 Aug 19 14:01 Godot_v4.7.2-stable_mono_linux_arm64.zip
-rw-r--r--. 1 akien akien  106024676 Aug 19 14:01 Godot_v4.7.2-stable_mono_linux_x86_32.zip
-rw-r--r--. 1 akien akien  105799045 Aug 19 14:00 Godot_v4.7.2-stable_mono_linux_x86_64.zip
-rw-r--r--. 1 akien akien  195951460 Aug 19 14:08 Godot_v4.7.2-stable_mono_macos.universal.zip
-rw-r--r--. 1 akien akien  116185669 Aug 19 14:05 Godot_v4.7.2-stable_mono_win32.zip
-rw-r--r--. 1 akien akien  114330231 Aug 19 14:04 Godot_v4.7.2-stable_mono_win64.zip
-rw-r--r--. 1 akien akien  112711157 Aug 19 14:06 Godot_v4.7.2-stable_mono_windows_arm64.zip
-rw-r--r--. 1 akien akien       1719 Aug 19 14:12 SHA512-SUMS.txt
```

</details>

This isn't just unattended CI time, as @Repiteo needs to manually run `publish-release.sh` (which also takes a while) after `build-release.sh`, so the time between the two matters in a work day.

---

*Edit:* Better size comparison including only the zips we make with `build-release.sh`:

|Compression|Size|
|---|---|
|`zip -9`|3.955 GiB|
|`7z -tzip -mx9`|3.873 GiB|

So that's now a 2% size reduction which is a bit better. But still not worth the change IMO.

akien-mga:
Force pushed a rebase on latest `main` with the commit amended to fix some issues I pointed out earlier. The results in my previous comment were obtained with this commit.

bruvzg:
> Compress the templates .tpz with a better format like zstd, which we support in the engine. But I don't know if the method we use to extract specific templates from the .tpz with the new Export Templates Manager can work with non-zip formats.

If it's `zip` container using `zstd` compression method, everything we use to download/extract specific parts should probably work, but it's not supported by currently used version of minizip (minizip-ng seems to support it). But it might also reduce size saving comparing to `tar.zstd` (since like regular zip it will compress individual files, not the whole archive).

Calinou:
`.tar.zst` should be usable as a baseline for Linux editor downloads, but I don't think it's available out of the box on macOS (at least in Finder).

On Windows, it's only usable out of the box on Windows 11 (which has libarchive integration with Zstandard enabled), not on Windows 10.

akien-mga:
> If it's `zip` container using `zstd` compression method, everything we use to download/extract specific parts should probably work, but it's not supported by currently used version of minizip (minizip-ng seems to support it). But it might also reduce size saving comparing to `tar.zstd` (since like regular zip it will compress individual files, not the whole archive).

That sounds quite interesting. Do you know of command line tools that can make zstd-compressed zips?
And would Windows 10 and 11 be able to decompress those zips? That could be a good tradeoff between better compression and a widely supported archive container format (unlike tar).

Upgrading our minizip to minizip-ng would be worth doing if it enables us to use this format.

As for size saving, I can do some tests. For Linux/Windows archives (exes and debug symbols) it's most single files or at most two files with the `_console.exe`, so it should be pretty good. And for macOS aside from the executable most other files are tiny in the `.app`. The `.dSYM` folder may have the worst tradeoff.

*Edit:* Did some testing on Linux and it seems support for compression zips with zstd isn't widespread, despite having been added to the standard in 2020. Clankers also claim that decompression support would also be limited to recent 7z and WinZip, but isn't supported by Windows Explore, macOS's archiving tool, or plain `unzip`. So probably not an improvement over using `.tar.zst`.

Calinou:
> Do you know of command line tools that can make zstd-compressed zips?

Official 7z lacks Zstandard support, but according to [this blog post](https://nickb.dev/blog/there-and-back-again-with-zstd-zips/), this fork of p7zip can do it: https://github.com/p7zip-project/p7zip

But unfortunately, Windows 11's built-in extraction tool doesn't handle anything other than Deflate or Store in a ZIP.

PR #87, opened 2023-10-04
Publish Android library from the main build-release script

YuriSizov:
This is a requirement for all current versions of Godot, both stable and pre-releases. I unified this new option with the old one to publish nuget packages, as it all seems related. (Old `-n` options is kept for compatibility.)

I haven't tested this, not really sure how, but I think it should be pretty straight-forward.

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -1,6 +1,7 @@
 #!/bin/bash
 
 set -e
+export basedir=$(pwd)
 
 # Config
 
@@ -104,6 +105,21 @@ publish_nuget_packages() {
   done
 }
 
+can_publish_maven=0
+if [ ! -d "${basedir}/deps/keystore" ]; then
+  echo "Disabling Android library publishing as ${basedir}/deps/keystore doesn't exist."
+else
+  can_publish_maven=1
+fi
+
+publish_maven_library() {
+  if [ $can_publish_maven == 0 ]; then
+    return
+  fi
+  # FIXME: Might be worth reworking the script to make it all sudo-safe and use appropriate users throughout.
+  sudo sh build-android/upload-mavencentral.sh
+}
+
 godot_version=""
 templates_version=""
 do_cleanup=1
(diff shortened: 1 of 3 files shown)


build-release.sh
```diff
@@ -104,23 +105,39 @@ publish_nuget_packages() {
...
+  echo "Disabling Android library publishing as ${basedir}/deps/keystore doesn't exist."
+else
+  can_publish_maven=1
+fi
+
+publish_maven_library() {
+  if [ $can_publish_maven == 0 ]; then
+    return
+  fi
+  sh build-android/upload-mavencentral.sh
```
akien-mga, line 119: So the main reason why I kept this as a separate script, is that this script requires `sudo`, but the `build-release.sh` isn't run with `sudo`.

IIRC running it with sudo fails as it's using the wrong user for connecting to HP's Mac over ssh. Can likely be fixed, but needs some experimentation.

YuriSizov: Ah, right, I remember. So adding `sudo` to this line should be sufficient then? I don't think this would elevate the whole script if we do it this way, would it?

build-release.sh
```diff
@@ -104,23 +105,40 @@ publish_nuget_packages() {
...
   case "$opt" in
   h|\?)
     echo "Usage: $0 [OPTIONS...]"
     echo
     echo "  -v godot version (e.g: 3.2-stable) [mandatory]"
     echo "  -t templates version (e.g. 3.2.stable) [mandatory]"
     echo "  -b build target: all|classical|mono|none (default: all)"
-    echo "  -n publish nuget packages (default: false)"
+    echo "  -p publish target: all|nuget|maven|none (default: none)"
+    echo "  -n publish nuget packages (default: false; legacy option, use -p instead)"
```
akien-mga, line 141: I think we can drop `-n`, this is only used by us, we can adapt to the new syntax (as long as using `-n` properly errors out, so we don't do mistakes when changing branches, etc.).

akien-mga (approved):

akien-mga:
I havent tested (should be done on a build we intend to upload, e.g. 4.2 beta 1), but this looks fine.

The use of `sudo` within the script means that when it reaches this point, it will ask for password, so we need to monitor it to finish running `build-release.sh` successfully. That should be fine for now, I may improve things later on (e.g. making the whole script `sudo` safe so we run it upfront with `sudo`).

build-release.sh
YuriSizov: Looks like it should error out, so removed!

YuriSizov:
Changes done!

akien-mga:
Thanks!

PR #89, opened 2023-10-24
Update scripts for new Fedora 39 based images for 4.2-beta3 and later

akien-mga:
Follow-up to https://github.com/godotengine/build-containers/pull/128.

Also reverts #88 since these new images include JDK 17, and the closure compiler issue was fixed upstream.

And removes the manual install of gettext which is now also part of the images.

X11 libs should not be needed to generate the Mono glue anymore (they've been unnecessary for a while already).


akien-mga:
> Draft until I confirm that 4.2-beta3 builds fine with this.

Tested successfully.

PR #90, opened 2023-11-01
Linux: Update to new SDK with GCC 13, add ARM64 and ARM32

akien-mga:
- Depends on https://github.com/godotengine/build-containers/pull/131.
- Supersedes #52.


akien-mga:
I'll update to remove the `strip` calls, we already strip binaries when building, and x86_64 strip can't strip arm binaries.

PR #91, opened 2023-11-23
Linux: Disable LTO for ARM64/ARM32, it crashes on Raspberry Pi OS

akien-mga:
Needs further investigation and reporting the bug upstream.

Technically I've only tested it to be an issue on ARM64 on a Raspberry Pi 5, but I'm disabling it for ARM32 too to be safe.

@hiulit reported that Ubuntu 23.10 seemed to work fine with our LTO'd binaries, so looking at the differences between RPiOS and Ubuntu 23.10 might help pinpoint what causes the issue.

Might be GCC or glibc bug?

### DIFF
--- a/build-linux/build.sh
+++ b/build-linux/build.sh
@@ -7,6 +7,7 @@ set -e
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
 export OPTIONS="production=yes"
 export OPTIONS_MONO="module_mono_enabled=yes"
+export OPTIONS_ARM="lto=none"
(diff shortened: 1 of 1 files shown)


akien-mga:
Tested successfully.

akien-mga:
Reverted by ceadb16e22f6c829c54997838885a1b6b84f12ec, this didn't actually fix the crash. It was fixed by https://github.com/godotengine/build-containers/pull/134. I tested a build with LTO and a Linux container from that PR, it works fine on RPi 5.

PR #92, opened 2024-01-04
[Windows] Enable D3D12 support.

bruvzg:
Partially fixes https://github.com/godotengine/godot/issues/86490

It won't download and package proprietary `dxil.dll` with the editor/templates, but it's loaded dynamically so it won't prevent the editor to work with other renderers and can be added without building the engine. Also, might work if Windows is in developer mode, since library is used only for shader signing.

### DIFF
--- a/build-windows/build.sh
+++ b/build-windows/build.sh
@@ -5,7 +5,7 @@ set -e
 # Config
 
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
-export OPTIONS="production=yes angle_libs=/root/angle"
+export OPTIONS="production=yes angle_libs=/root/angle mesa_libs=/root/mesa d3d12=yes"
 export OPTIONS_MONO="module_mono_enabled=yes"
 export TERM=xterm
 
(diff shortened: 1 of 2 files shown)


build-windows/build.sh
```diff
@@ -5,7 +5,7 @@ set -e
 # Config
 
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
-export OPTIONS="production=yes angle_libs=/root/angle"
+export OPTIONS="production=yes angle_libs=/root/angle mesa_libs=/root/mesa d3d12=yes"
```
akien-mga, line 8: Don't we need `d3d12=yes` too? Or are we going to make it the default?

bruvzg: > Don't we need d3d12=yes too? 

We do, and it is there (last param). We do not need `dxc_path` to build, it's only used to copy dll (headers and in the `thirdparty`).

akien-mga: Oh right I missed it :D

akien-mga (approved):

akien-mga:
Thanks!

PR #93, opened 2024-01-17
[3.x] Update all to new toolchains, add Linux arm64 and arm32 builds

akien-mga:
Backport of #89 and #90, in sync with https://github.com/godotengine/build-containers/pull/135.

In the 3.x branch, the situation with `arch` and `bits` is very brittle, so we only add the explicit `arch` argument for the arm32/arm64 builds. x86_32 still relies on `bits=32`.

This would all be worth refactoring upstream like we did for 4.0, but it's a major undertaking and breaking change, which I'd prefer to avoid in 3.6.

For Linux builds, we move the `strip` calls to the link stage, as this
needs to be done with the arch-appropriate `strip` binary, so it's easier
done there. In `master`, we now let the compiler strip automatically
during the build if no debug symbols are requested, but this change
wasn't backported to 3.x.

---

Tested successfully for almost all platforms, I'm just battling some final issues with Linux arm32 (https://github.com/godotengine/godot/pull/87258, and it looks like our buildroot will need to change its config to enable NEON intrinsics).

This is meant to be used with `3.x-f39-mono-6.12.0.198` containers built from https://github.com/godotengine/build-containers/pull/135.

### DIFF
--- a/build-ios/build.sh
+++ b/build-ios/build.sh
@@ -11,7 +11,7 @@ export OPTIONS="production=yes lto=none"
 export OPTIONS_MONO="module_mono_enabled=yes mono_static=yes"
 export TERM=xterm
 
-export IOS_SDK="15.4"
+export IOS_SDK="17.0"
 export IOS_LIPO="/root/ioscross/arm64/bin/arm-apple-darwin11-lipo"
 
 rm -rf godot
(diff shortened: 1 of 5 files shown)


akien-mga:
Tested successfully.

PR #102, opened 2024-12-06
Add ICU data to export templates.

bruvzg:
Part of https://github.com/godotengine/godot/pull/100104


bruvzg:
> can you add a check if the source exists, and just print a warning if the source .icu data doesn't exist

Done.

PR #106, opened 2025-02-28
Update to latest version of Swappy

darksylinc:
See godotengine/godot#103409

Fixes godotengine/godot#103294

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -184,7 +184,7 @@ if [ ! -d "deps/swappy" ]; then
   echo "Missing Swappy libraries, downloading them."
   mkdir -p deps/swappy
   pushd deps/swappy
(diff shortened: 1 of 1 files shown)


akien-mga:
@darksylinc Would you be up for moving the `godot-swappy` repo under the `godotengine` organization? It would be a bit cleaner that we download binaries from a repo we host officially. You'd stay maintainer on that repo obviously.

I sent you an invite to the @godotengine org which should then let you initiate a transfer of ownership, if you're willing. Then I might need to re-add you as maintainer on that repo as I believe GH might remove it in the process.

darksylinc:
[It is done](https://github.com/godotengine/godot-swappy).

darksylinc:
Ok, changed reference from darksylinc/ to godotengine/

akien-mga (approved):

akien-mga:
Thanks!

PR #109, opened 2025-03-20
Add accesskit-c to the build setup.

bruvzg:


build.sh
```diff
@@ -148,6 +148,16 @@ if [ ! -d "deps/moltenvk" ]; then
   popd
 fi
 
+# accesskit-c for Windows, macOS and Linux
+if [ ! -d "deps/accesskit" ]; then
+  echo "Missing accesskit, downloading it."
+  mkdir -p deps/accesskit
+  pushd deps/accesskit
+  curl -L -o accesskit.zip https://github.com/bruvzg/godot-accesskit-c-static/releases/download/0.15.1/accesskit-c-0.15.1.zip
```
bruvzg, line 156: I'll transfer the custom build repo and change this later.

build-linux/build.sh
```diff
@@ -7,6 +7,7 @@ set -e
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no"
 export OPTIONS="production=yes"
 export OPTIONS_MONO="module_mono_enabled=yes"
+export OPTIONS_ACKIT="accesskit_sdk_path=/root/accesskit/accesskit-c-0.15.1"
```
akien-mga, line 10: Can probably be in the main `OPTIONS` like for macOS and Windows?

I'd also suggest renaming the folder to remove the version in the main script, so we only have one location to update when changing version.

bruvzg: Added it separately since there's no arm Linux libs in official build, and I have not configured cross build for it.

> I'd also suggest renaming the folder to remove the version in the main script, so we only have one location to update when changing version.

Done.

bruvzg: > Added it separately since there's no arm Linux libs in official build, and I have not configured cross build for it.

Added arm Linux build configs to the https://github.com/bruvzg/godot-accesskit-c-static and moved it `OPTIONS`.

akien-mga (approved):

PR #110, opened 2025-03-26
publish-release: Add support to upload stable builds to GH, Steam, EGS and itch.io

akien-mga:
This is hacked together but seems to work, tested with 4.4.1-stable.

More work would be good eventually to rely less on some adhoc configuration of Steam/EGS/godot-builds paths on our build server. We should move those configs to a Git repository (this one, or a separate one, possibly private as only us should use it), but this should be done carefully as there are some passwords involved (which should be moved to `config.sh`).

For now, this should be a good start though to automate some of the process.

Worth noting, for a stable release, `./publish-release.sh -l` can be used to mark the release as latest - this should only be used for the most recent supported release (e.g. 4.4.1 now, but a future 4.3.1 shouldn't be marked as stable). This controls which version of the Web editor is the default one, and whether to upload builds to EGS and itch.io (where we only maintain one version). We upload all versions to Steam as we have version branches there.

**Before running the script**, the version bump commit and **Git tag** should be pushed to godotengine/godot. Not made into a GitHub Release though, this is done by the script.

**After running the script**, builds still need to be made live manually on Steam and EGS in their respective backends. On itch it's automatically live.

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -3,6 +3,9 @@
 set -e
 export basedir=$(pwd)
 
+# Log output to a file automatically.
+exec > >(tee -a "out/logs/build-release") 2>&1
+
 # Config
 
 # For signing keystore and password.
@@ -266,13 +269,6 @@ if [ "${build_classical}" == "1" ]; then
   zip -q -9 "${reldir}/${binname}.zip" ${binname} ${wrpname}
   rm ${binname} ${wrpname}
 
-  if [ -d out/windows/steam ]; then
-    cp out/windows/steam/godot.windows.editor.x86_64.exe ${steamdir}/godot.windows.opt.tools.64.exe
-    cp out/windows/steam/godot.windows.editor.x86_32.exe ${steamdir}/godot.windows.opt.tools.32.exe
-    sign_windows ${steamdir}/godot.windows.opt.tools.64.exe
-    sign_windows ${steamdir}/godot.windows.opt.tools.32.exe
-  fi
-
   # Templates
   cp out/windows/x86_64/templates/godot.windows.template_release.x86_64.exe ${templatesdir}/windows_release_x86_64.exe
   cp out/windows/x86_64/templates/godot.windows.template_debug.x86_64.exe ${templatesdir}/windows_debug_x86_64.exe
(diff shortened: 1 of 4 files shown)


build-release.sh
```diff
@@ -312,6 +308,20 @@ if [ "${build_classical}" == "1" ]; then
...
+    cp out/windows/steam/godot.windows.editor.x86_64.exe ${steamdir}/godot.windows.opt.tools.64.exe
+    cp out/windows/steam/godot.windows.editor.x86_32.exe ${steamdir}/godot.windows.opt.tools.32.exe
+    sign_windows ${steamdir}/godot.windows.opt.tools.64.exe
+    sign_windows ${steamdir}/godot.windows.opt.tools.32.exe
+    unzip ${reldir}/${godot_basename}_linux.x86_64.zip -d ${steamdir}/
+    unzip ${reldir}/${godot_basename}_linux.x86_32.zip -d ${steamdir}/
+    mv ${steamdir}/{${godot_basename}_linux.x86_64,godot.x11.opt.tools.64}
+    mv ${steamdir}/{${godot_basename}_linux.x86_32,godot.x11.opt.tools.32}
+    unzip ${reldir}/${godot_basename}_macos.universal -d ${steamdir}/
+  fi
```
akien-mga, line 323: Here we're preparing the Steam binaries to be signed (for Windows, it's a custom Steam-specific build) and with the expected names (for Linux). This is just to make it easier to copy the binaries in `publish-release.sh` directly.

publish-release.sh
```diff
@@ -46,26 +57,108 @@ case "$choice" in
...
+  read -r RELEASE_TITLE RELEASE_DESC <<< "$release_info"
+
+  gh release create ${godot_version} --repo godotengine/godot --title "$RELEASE_TITLE" --notes "$RELEASE_DESC"
+  gh release upload ${godot_version} ${reldir}/[Gg]* ${reldir}/mono/[Gg]*
+  # Concatenate SHA sums.
+  cp ${reldir}/SHA512-SUMS.txt .
+  cat ${reldir}/mono/SHA512-SUMS.txt >> SHA512-SUMS.txt
+  gh release upload ${godot_version} SHA512-SUMS.txt
+  rm SHA512-SUMS.txt
+  popd
```
akien-mga, line 125: This copies the release info from godotengine/godot-builds and uploads it anew to godotengine/godot.

As we don't give @GodotBuilder write permissions on godotengine/godot for security, this requires one of us to get a PAT for upload. I suggest making one with 1 day expiry each time this is needed, which is why the script asks for it at the start (so it's only pasted in the terminal and not hardcoded anywhere).

publish-release.sh
```diff
@@ -91,8 +184,40 @@ else
...
+
+        echo "Command failed. Retrying in $DELAY seconds..."
+        sleep $DELAY
+        ((attempt++))
+        DELAY=$((DELAY * 2))  # Exponential backoff
+    done
+
+    echo "❌ Command failed after $MAX_RETRIES attempts."
+    return 1
+}
```
akien-mga, line 209: As the web editor upload SSH connection often gets aborted by the remote server, I finally put it some retry logic. Courtesy of ChatGPT /o\

I still think this is super hacky and could be done better, but if it does the work for now, should be good.

publish-release.sh
akien-mga: I need to recheck this, as it didn't seem to work fully in https://github.com/godotengine/godot/releases/tag/4.4.1-stable - I've now copied the text from the godot-builds release manually, but initially the description was empty.

Also, we should handle the `${latest_stable}` flag here too to make sure that only the latest stable gets flagged as latest release on GitHub.

akien-mga: > I need to recheck this, as it didn't seem to work fully in https://github.com/godotengine/godot/releases/tag/4.4.1-stable - I've now copied the text from the godot-builds release manually, but initially the description was empty.

Found a fix using `jq`.

> Also, we should handle the `${latest_stable}` flag here too to make sure that only the latest stable gets flagged as latest release on GitHub.

Seems like we might be able to rely on GH's default logic here:
```
$ gh release create --help | grep Latest
      --latest                       Mark this release as "Latest" (default: automatic based on date and version)
```
Since the default is to guess, and there's no way to say "not latest", I guess we don't really have much options :D

akien-mga:
For the record I've co-developed this alongside publishing 4.4.1-stable so it's tested, but oftentimes in chunks and not the full process all at once with no manual tweaks.

So the real test of fire will be 4.4.2-stable or 4.3.1-stable, whichever comes first.

PR #115, opened 2025-06-05
Apple: add visionOS build support; unify build container

stuartcarnie:
Depends:

- https://github.com/godotengine/build-containers/pull/154

Adds support for building visionOS and tvOS templates

### DIFF
--- a/build-ios/build.sh
+++ b/build-ios/build.sh
@@ -12,7 +12,10 @@ export OPTIONS_MONO="module_mono_enabled=yes"
 export TERM=xterm
 
 export IOS_SDK="18.5"
-export IOS_LIPO="/root/ioscross/arm64/bin/arm-apple-darwin11-lipo"
+export IOS_DEVICE="IOS_SDK_PATH=/root/SDKs/iPhoneOS${IOS_SDK}.sdk"
+export IOS_SIMULATOR="IOS_SDK_PATH=/root/SDKs/iPhoneSimulator${IOS_SDK}.sdk simulator=yes"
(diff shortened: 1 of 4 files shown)


akien-mga (approved):

akien-mga:
Tested all the build process successfully!

What's missing now is just confirming that the builds are actually usable to export to visionOS (and iOS as we changed the setup a bit so we should validate it).

- Standard:
  * macOS editor: https://github.com/akien-mga/godot/releases/download/_tmp/Godot_v4.5-dev-f42_macos.universal.zip
  * Export templates (iOS, visionOS, macOS): https://github.com/akien-mga/godot/releases/download/_tmp/Godot_v4.5-dev-f42_export_templates.tpz
- Mono (C#):
  * macOS editor: https://github.com/akien-mga/godot/releases/download/_tmp/Godot_v4.5-dev-f42_mono_macos.universal.zip
  * Export templates: https://github.com/akien-mga/godot/releases/download/_tmp/Godot_v4.5-dev-f42_mono_export_templates.tpz

I'm curious to also get test reports for the C# support for visionOS. We haven't done any bespoke setup for it, does it Just Work™? (I doubt it, and I see we have quite a bit of iOS references in `modules/mono` but no visionOS, so this likely needs to be added.)

So we should probably disable the Mono builds for visionOS for now.

stuartcarnie:
@akien-mga I have not tested C#, but I suspect you are right, so disabling for now is probably a good idea. I can explore that next.

stuartcarnie:
I had a look in `modules/mono`, and I saw `APPLE_EMBEDDED_ENABLED`, so perhaps it does Just Work™? The architectures are both arm64.

akien-mga:
Thanks! Amazing work :tada:

PR #118, opened 2025-06-18
Windows: Change code signing process to match new key

akien-mga:


build-release.sh
```diff
@@ -12,17 +12,17 @@ exec > >(tee -a "out/logs/build-release") 2>&1
...
-  echo "Disabling Windows binary signing as config.sh does not define the required data (SIGN_KEYSTORE, SIGN_PASSWORD), or osslsigncode can't be found in PATH."
+  echo "Disabling Windows binary signing as config.sh does not define the required data (WINDOWS_SIGN_NAME, WINDOWS_SIGN_URL), or osslsigncode can't be found in PATH."
 fi
 
 sign_windows() {
   if [ $can_sign_windows == 0 ]; then
     return
   fi
-  osslsigncode sign -pkcs12 ${SIGN_KEYSTORE} -pass "${SIGN_PASSWORD}" -n "${SIGN_NAME}" -i "${SIGN_URL}" -t http://timestamp.comodoca.com -in $1 -out $1-signed
+  P11_KIT_SERVER_ADDRESS=unix:path=/run/p11-kit/p11kit.sock osslsigncode sign -pkcs11module /usr/lib64/pkcs11/p11-kit-client.so -pkcs11cert 'pkcs11:model=SimplySign%20C' -key 'pkcs11:model=SimplySign%20C' -t http://time.certum.pl/ -n "${WINDOWS_SIGN_NAME}" -i "${WINDOWS_SIGN_URL}" -in $1 -out $1-signed
```
akien-mga, line 25: @hpvb Is there some of it that should be moved to `config.sh` so third-parties using our scripts can also do their own signing with the same system?

I don't see any "secret" so I guess the actual key registration happens server side and it's accessed though p11-kit? Should we document how this works?

hpvb: Yeah, there's nothing secret about that, now that the container is ready (https://github.com/hpvb/certum-container) we could probably just document it. But that would work only for Certum, not necessarily other CAs

Digicert, globalsign, and sertigo all have different systems, although the container could probably be adapted for other CAs. There's no way for me to do that tho without buying a bunch more certs.

PR #112, opened 2025-04-20
Setup Android debug symbols

m4gr3d:
Build scripts counterpart to https://github.com/godotengine/godot/pull/105605.

This PR updates the Android build process to generate native debug symbols. This is done by setting `debug_symbols=yes` for the `scons` build, and letting `gradle` strips out the generated `so` artifacts. As part of the stripping process, `gradle` will generate a separate debug symbols.

PR #120, opened 2025-06-26
Android: Fix native debug symbols generation

m4gr3d:
Follow-up to https://github.com/godotengine/godot-build-scripts/pull/112, depends on https://github.com/godotengine/godot/pull/107998.

The native debug symbols file generated by `gradle` were invalid and could not be used to symbolicate native crashes. With https://github.com/godotengine/godot/pull/107998, the process has been updated to use the `.so` files generated by `scons` instead.

As a side effect, the `aab` no longer contains the native debug symbols file (although this means the file will need to be manually uploaded to the Play store) and is much smaller: `1.1GB` -> `280MB`

### DIFF
--- a/build-android/build.sh
+++ b/build-android/build.sh
@@ -4,8 +4,8 @@ set -e
 
 # Config
 
-# Debug symbols are enabled for the Android builds. Gradle will strip them out of
-# the final artifacts and generate a separate debug symbols file.
+# Debug symbols are enabled for the Android builds so we can generate a separate debug symbols file. 
+# Gradle will strip them out of the final artifacts.
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no redirect_build_objects=no"
 export OPTIONS="production=yes debug_symbols=yes"
 export OPTIONS_MONO="module_mono_enabled=yes"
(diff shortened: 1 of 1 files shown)


build-android/build.sh
```diff
@@ -40,7 +40,7 @@ if [ "${CLASSICAL}" == "1" ]; then
   $SCONS platform=android arch=arm32 $OPTIONS target=editor store_release=${store_release}
   $SCONS platform=android arch=arm64 $OPTIONS target=editor store_release=${store_release}
   $SCONS platform=android arch=x86_32 $OPTIONS target=editor store_release=${store_release}
-  $SCONS platform=android arch=x86_64 $OPTIONS target=editor store_release=${store_release}
+  $SCONS platform=android arch=x86_64 $OPTIONS target=editor store_release=${store_release} separate_debug_symbols=yes # Generate native debug symbols.
```
Calinou, line 43: Is it intended for `separate_debug_symbols=yes` to be only present for the last architecture being built? Does it create debug symbols for all architectures at once? I thought symbols were architecture-specific.

m4gr3d: It's intended to be last because all it does is zip the directory `platform/android/java/lib/libs/tools/release/` containing the `.so` files generated by each architecture:
- `platform/android/java/lib/libs/tools/release/arm64-v8a` for arm64
- `platform/android/java/lib/libs/tools/release/armeabi-v7a` for arm32
- ...

syntaxerror247 (approved):

Repiteo (approved):

PR #156, opened 2026-04-11
Apple: visionOS and tvOS builds now supported; update to 26.4 SDKs

stuartcarnie:
Depends on:

- https://github.com/godotengine/build-containers/pull/162

Changes in this PR:

- Adds tvOS support via a new `build-tvos/build.sh`, mirroring `build-ios/build.sh` but arm64-device only (no simulator) for both Classical and Mono templates. Outputs land under `out/templates/libgodot.tvos*.a` and `out/templates-mono/libgodot.tvos*.a`.
- Bumps the iOS and visionOS SDKs from 26.1 / 26.0 to **26.4**, matching the Xcode 26.4 SDKs pulled into the `godot-appleembedded` image.
- Bumps the Swift toolchain used by the iOS and visionOS builds from 6.2.x to **6.3.0** (Apple SDK 26.4 shipped with swiftlang-6.3.0.123.x, so the OSS Swift 6.3 release is the closest marketing-version match).

Paired with the [`build-containers` PR ](https://github.com/godotengine/build-containers/pull/162) that adds the Swift 6.3 toolchain to the `godot-appleembedded` image and applies the necessary visionOS SDK patches.

### DIFF
--- a/build-ios/build.sh
+++ b/build-ios/build.sh
@@ -7,12 +7,12 @@ set -e
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no redirect_build_objects=no"
 # Keep LTO disabled for iOS - it works but it makes linking apps on deploy very slow,
 # which is seen as a regression in the current workflow.
-export OPTIONS="production=yes use_lto=no SWIFT_FRONTEND=/root/.local/share/swiftly/toolchains/6.2.1/usr/bin/swift-frontend"
+export OPTIONS="production=yes use_lto=no SWIFT_FRONTEND=/root/.local/share/swiftly/toolchains/6.3.0/usr/bin/swift-frontend"
 export OPTIONS_MONO="module_mono_enabled=yes"
 export OPTIONS_DOTNET="module_dotnet_enabled=yes"
 export TERM=xterm
 
-export IOS_SDK="26.1"
+export IOS_SDK="26.4"
 export IOS_DEVICE="IOS_SDK_PATH=/root/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS${IOS_SDK}.sdk"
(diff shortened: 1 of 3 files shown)


build-ios/build.sh
```diff
@@ -7,12 +7,12 @@ set -e
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no redirect_build_objects=no"
 # Keep LTO disabled for iOS - it works but it makes linking apps on deploy very slow,
 # which is seen as a regression in the current workflow.
-export OPTIONS="production=yes use_lto=no SWIFT_FRONTEND=/root/.local/share/swiftly/toolchains/6.2.1/usr/bin/swift-frontend"
+export OPTIONS="production=yes use_lto=no SWIFT_FRONTEND=/root/.local/share/swiftly/toolchains/6.3.0/usr/bin/swift-frontend"
```
akien-mga, line 10: I think we should add a `SWIFT_VERSION` env to `Dockerfile.osx` so we can reuse it here and in the other places.

build-ios/build.sh
```diff
@@ -7,12 +7,12 @@ set -e
...
 export OPTIONS_MONO="module_mono_enabled=yes"
 export OPTIONS_DOTNET="module_dotnet_enabled=yes"
 export TERM=xterm
 
-export IOS_SDK="26.1"
+export IOS_SDK="26.4"
```
akien-mga, line 15: I could tackle this in a follow-up, but we could also expose this as an env variable so we don't need to adjust build scripts every time we update the SDKs.

akien-mga (commented):
You also need to update the Swift version in `build-macos/build.sh`.

And likewise there we could move `darwin25.1` to an env variable in the build container so we don't need to do updates like this every time we update SDKs.

stuartcarnie:
Closed in favour of the New Improved™:

- #159

PR #157, opened 2026-04-11
Enable perfetto for debug / release templates

m4gr3d:
Follow-up to https://github.com/godotengine/godot/pull/118401

### DIFF
--- a/build-android/build.sh
+++ b/build-android/build.sh
@@ -19,6 +19,8 @@ prepare_source() {
   cd godot
   tar xf /root/godot.tar.gz --strip-components=1
   cp -rf /root/swappy/* thirdparty/swappy-frame-pacing/
+  mkdir -p thirdparty/perfetto
+  cp -rf /root/perfetto/* thirdparty/perfetto/
 }
 
 # Environment variables and keystore needed for signing store editor build,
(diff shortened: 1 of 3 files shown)


m4gr3d:
@akien-mga @Repiteo Following my comments in https://github.com/godotengine/godot/pull/118401#issuecomment-4232262214, I think there are at least three ways we can go with this proposal:
- Keeping with the current status, i.e: not enabling perfetto on the Android production templates we provide, so we would keep providing the following Android production templates:
  - release template (no perfetto)
  - debug template (no perfetto)
- Enabling perfetto only for debug production templates, so we would provide the following Android production templates:
  - release template (no perfetto)
  - debug template with perfetto
- Enabling perfetto for both debug and release production templates, but keeping them as a separate download targets, so we would provide the following Android production templates:
  - release template (no perfetto)
  - debug template (no perfetto)
  - release template with perfetto
  - debug template with perfetto

dsnopek:
The tricky thing is that profiling really _should_ be done with release builds. Sometimes a thing can be slow on a debug build, and just magically get faster when running on a release build, and so if folks are primarily profiling debug builds, they may end up optimizing things that don't really need to be optimized.

So, while it would be "safe" to enable Perfetto in debug builds by default (given that Perfetto will have some performance cost), it isn't nearly as useful as having Perfetto-enabled release builds available (but we don't necessarily want that as the default because of the performance cost)

Repiteo:
If we're opening the floodgates to perfetto in templates, shouldn't it be enabled across _all_ platforms? Given there's already been a push for separating/tidying build server outputs, this would be another reason to dig into that even further

m4gr3d:
> The tricky thing is that profiling really _should_ be done with release builds. Sometimes a thing can be slow on a debug build, and just magically get faster when running on a release build, and so if folks are primarily profiling debug builds, they may end up optimizing things that don't really need to be optimized.
> 
> So, while it would be "safe" to enable Perfetto in debug builds by default (given that Perfetto will have some performance cost), it isn't nearly as useful as having Perfetto-enabled release builds available (but we don't necessarily want that as the default because of the performance cost)

Hence the third option (you can disregard the current PR changes until we land on an approach among [those proposed](https://github.com/godotengine/godot-build-scripts/pull/157#issuecomment-4232319041)):

> Enabling perfetto for both debug and release production templates, but keeping them as a separate download targets

I'm leaning toward this approach as it doesn't incur any binary size or performance penalty for those not interested in profiling their game, but provide an avenue for those interested (so long as we provide good documentation for how to use the new templates).

m4gr3d:
> If we're opening the floodgates to perfetto in templates, shouldn't it be enabled across _all_ platforms? Given there's already been a push for separating/tidying build server outputs, this would be another reason to dig into that even further

perfetto support is [built-in the Android OS](https://perfetto.dev/docs/#why-would-you-use-perfetto-) so it makes sense to have it enabled on Android. It either [doesn't integrate as well on other platforms](https://perfetto.dev/docs/#why-would-you-strong-not-strong-use-perfetto-), or require more setup.

For example, on Apple platforms [instruments](https://developer.apple.com/tutorials/instruments) could be the default profiler instead.

m4gr3d:
I went with the 3rd option and creating separate debug and release perfetto templates. They are not packaged within the export templates, but are available for download from the build page.

Repiteo (approved):
Let's give it a shot!

PR #159, opened 2026-04-15
Apple: tvOS, visionOS support; use unified build container

stuartcarnie:
* Updated to use new build container, godot-apple
* Adds visionOS target
* Adds support for tvOS but does not compile yet as depends on:
    * https://github.com/godotengine/godot/pull/118332

Depends on:

- https://github.com/godotengine/build-containers/pull/163
- https://github.com/godotengine/godot/pull/121255

# Testing

- Built the macOS editor App
- Build the templates for iOS and visionOS
- Generated the tpz export archive
- Ran the editor on macOS and exported the Platformer 2D to iOS and visionOS and successfully compiled the binaries under Xcode.

### DIFF
--- a/build-ios/build.sh
+++ b/build-ios/build.sh
@@ -4,15 +4,20 @@ set -e
 
 # Config
 
+# Swift toolchain path comes from the image (godot-apple sets SWIFT_VERSION);
+# fall back to a sensible default if run against an older image.
+SWIFT_VERSION="${SWIFT_VERSION:-6.3.0}"
+SWIFT_FRONTEND="/root/.local/share/swiftly/toolchains/${SWIFT_VERSION}/usr/bin/swift-frontend"
+
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no redirect_build_objects=no"
 # Keep LTO disabled for iOS - it works but it makes linking apps on deploy very slow,
 # which is seen as a regression in the current workflow.
-export OPTIONS="production=yes use_lto=no SWIFT_FRONTEND=/root/.local/share/swiftly/toolchains/6.2.1/usr/bin/swift-frontend"
+export OPTIONS="production=yes use_lto=no SWIFT_FRONTEND=${SWIFT_FRONTEND}"
 export OPTIONS_MONO="module_mono_enabled=yes"
 export OPTIONS_DOTNET="module_dotnet_enabled=yes"
 export TERM=xterm
 
-export IOS_SDK="26.1"
+export IOS_SDK="26.4"
 export IOS_DEVICE="IOS_SDK_PATH=/root/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS${IOS_SDK}.sdk"
 export IOS_SIMULATOR="IOS_SDK_PATH=/root/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator${IOS_SDK}.sdk simulator=yes"
 export APPLE_TARGET_ARM64="APPLE_TOOLCHAIN_PATH=/root/ioscross/arm64 apple_target_triple=arm-apple-darwin11-"
--- a/build-ios/build.sh
+++ b/build-ios/build.sh
@@ -6,7 +6,7 @@ set -e
 
 # Swift toolchain path comes from the image (godot-apple sets SWIFT_VERSION);
 # fall back to a sensible default if run against an older image.
-SWIFT_VERSION="${SWIFT_VERSION:-6.3.0}"
+SWIFT_VERSION="${SWIFT_VERSION:-6.3.2}"
 SWIFT_FRONTEND="/root/.local/share/swiftly/toolchains/${SWIFT_VERSION}/usr/bin/swift-frontend"
 
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no redirect_build_objects=no"
@@ -17,7 +17,7 @@ export OPTIONS_MONO="module_mono_enabled=yes"
 export OPTIONS_DOTNET="module_dotnet_enabled=yes"
 export TERM=xterm
 
-export IOS_SDK="26.4"
+export IOS_SDK="26.5"
 export IOS_DEVICE="IOS_SDK_PATH=/root/Xcode.app/Contents/Developer/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS${IOS_SDK}.sdk"
 export IOS_SIMULATOR="IOS_SDK_PATH=/root/Xcode.app/Contents/Developer/Platforms/iPhoneSimulator.platform/Developer/SDKs/iPhoneSimulator${IOS_SDK}.sdk simulator=yes"
 export APPLE_TARGET_ARM64="APPLE_TOOLCHAIN_PATH=/root/ioscross/arm64 apple_target_triple=arm-apple-darwin11-"
--- a/build-ios/build.sh
+++ b/build-ios/build.sh
@@ -7,16 +7,14 @@ set -e
 # Swift toolchain path comes from the image (godot-apple sets SWIFT_VERSION);
 # fall back to a sensible default if run against an older image.
 SWIFT_VERSION="${SWIFT_VERSION:-6.3.2}"
-# Swift is compiled via SWIFT_FRONTEND (swift-frontend) today; SWIFT_COMPILER
-# (swiftc) is passed too but not yet consumed by scons — ready for when the
-# visionOS PR switches setup_swift_builder to swiftc.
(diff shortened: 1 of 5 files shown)


build-macos/build.sh
```diff
@@ -4,12 +4,36 @@ set -e
...
+#   -L .../XcodeDefault.xctoolchain/usr/lib/clang/21/lib/darwin -lclang_rt.osx
+#                                — Apple's compiler-rt from Xcode, providing
+#                                  __isPlatformVersionAtLeast for
+#                                  `if (@available(...))` / `__builtin_available`.
+#                                  Apple's clang driver auto-links this on
+#                                  macOS; osxcross-style cross-builds have to
+#                                  wire it up explicitly.
+XCODE_CLANG_RT="/root/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/lib/clang/21/lib/darwin"
+EXTRA_LINKFLAGS="linkflags=-fuse-ld=lld -B /opt/darwin-tools/bin -L${XCODE_CLANG_RT} -lclang_rt.osx"
+
```
akien-mga: Can this be added in upstream Godot's `platform/macos/detect.py` in some way?

Those build scripts aren't meant to have a lot of custom code compared to just configuring the build with standard SCons options we provide.

build-macos/build.sh
```diff
@@ -4,12 +4,36 @@ set -e
...
 # Config
 
+# Swift toolchain path comes from the image (godot-apple sets SWIFT_VERSION);
+# fall back to a sensible default if run against an older image.
+SWIFT_VERSION="${SWIFT_VERSION:-6.3.2}"
+SWIFT_COMPILER="/root/.local/share/swiftly/toolchains/${SWIFT_VERSION}/usr/bin/swiftc"
+
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no redirect_build_objects=no"
-export OPTIONS="osxcross_sdk=darwin25.1 production=yes use_volk=no vulkan_sdk_path=/root/moltenvk angle_libs=/root/angle accesskit_sdk_path=/root/accesskit/accesskit-c SWIFT_FRONTEND=/root/.local/share/swiftly/toolchains/6.2.1/usr/bin/swift-frontend"
+export OPTIONS="osxcross_sdk=darwin25.5 production=yes use_volk=no vulkan_sdk_path=/root/moltenvk angle_libs=/root/angle accesskit_sdk_path=/root/accesskit/accesskit-c SWIFT_COMPILER=${SWIFT_COMPILER}"
```
akien-mga: As I suggested on chat, I think we should take the opportunity of the new apple container to add an env variable for the `darwin` SDK version so we don't need to hardcode it here.

build-macos/build.sh
stuartcarnie: I'd argue these flags are internal details specific to this build container version, so I'd rather not leak them out to the `scons` script. Otherwise every time I change / improve this container, I'd also have to change the scons scripts in Godot. There is already some leakage, but at least today it is fairly minimal. 

My longer-term goal is to keep chipping away at these internal hacks so they aren't necessary.

hpvb: You can put these in the actual container as pre-set environments. That would be far preferable over having it in the build scripts. The build scripts themselves shouldn't depend on very specific versions of stuff in the containers.

For mono we also used to do that. Just add these to the environment in the container that actually has these versions installed!

stuartcarnie: Good call, thanks @hpvb – will do that

build-macos/build.sh
stuartcarnie: This is going away with my latest improvements – osxcross is no longer needed, as we use the xtools linker (more modern) and the official Swift C/C++/Obj-C/Swift toolchain

build-macos/build.sh
stuartcarnie: Done!

build-macos/build.sh
stuartcarnie: This is now gone; we no longer need to pass this based on the new build containers

build-ios/build.sh
```diff
@@ -4,15 +4,22 @@ set -e
...
 # Keep LTO disabled for iOS - it works but it makes linking apps on deploy very slow,
 # which is seen as a regression in the current workflow.
-export OPTIONS="production=yes use_lto=no SWIFT_FRONTEND=/root/.local/share/swiftly/toolchains/6.2.1/usr/bin/swift-frontend"
+export OPTIONS="production=yes use_lto=no SWIFT_COMPILER=${SWIFT_COMPILER}"
 export OPTIONS_MONO="module_mono_enabled=yes"
 export OPTIONS_DOTNET="module_dotnet_enabled=yes"
 export TERM=xterm
 
-export IOS_SDK="26.1"
+export IOS_SDK="26.5"
```
akien-mga, line 22: We should expose this in the build container for convenience.

build-ios/build.sh
```diff
@@ -4,15 +4,22 @@ set -e
 
 # Config
 
+# Swift toolchain path comes from the image (godot-apple sets SWIFT_VERSION);
+# fall back to a sensible default if run against an older image.
+SWIFT_VERSION="${SWIFT_VERSION:-6.3.2}"
+# Swift sources are compiled with swiftc, passed via SWIFT_COMPILER
+# (SWIFT_FRONTEND remains a back-compat alias in Godot's detect.py).
+SWIFT_COMPILER="/root/.local/share/swiftly/toolchains/${SWIFT_VERSION}/usr/bin/swiftc"
```
akien-mga, line 12: Maybe we can expose this in the build container too?

build-macos/build.sh
```diff
@@ -4,12 +4,29 @@ set -e
...
 export SCONS="scons -j${NUM_CORES} verbose=yes warnings=no progress=no redirect_build_objects=no"
-export OPTIONS="osxcross_sdk=darwin25.1 production=yes use_volk=no vulkan_sdk_path=/root/moltenvk angle_libs=/root/angle accesskit_sdk_path=/root/accesskit/accesskit-c SWIFT_FRONTEND=/root/.local/share/swiftly/toolchains/6.2.1/usr/bin/swift-frontend"
+export OPTIONS="production=yes use_volk=no vulkan_sdk_path=/root/moltenvk angle_libs=/root/angle accesskit_sdk_path=/root/accesskit/accesskit-c SWIFT_COMPILER=${SWIFT_COMPILER} MACOS_SDK_PATH=${MACOS_SDK_PATH}"
 export OPTIONS_MONO="module_mono_enabled=yes"
 export OPTIONS_DOTNET="module_dotnet_enabled=yes"
 export TERM=xterm
 
+# Wrap the image-provided macOS link flags in the scons `linkflags=` option.
+EXTRA_LINKFLAGS="linkflags=${EXTRA_LINK_FLAGS}"
+
```
akien-mga, line 29: Would be better to put this in `OPTIONS` instead of a separate env variable which needs to be added to each scons call.

build-visionos/build.sh
```diff
@@ -4,16 +4,23 @@ set -e
...
 # which is seen as a regression in the current workflow.
 # Disable Vulkan and MoltenVK for visionOS - visionOS doesn't support MoltenVK.
-export OPTIONS="production=yes use_lto=no vulkan=no SWIFT_FRONTEND=/root/.local/share/swiftly/toolchains/6.2.0/usr/bin/swift-frontend"
+export OPTIONS="production=yes use_lto=no vulkan=no SWIFT_COMPILER=${SWIFT_COMPILER}"
 export OPTIONS_MONO="module_mono_enabled=yes"
 export OPTIONS_DOTNET="module_dotnet_enabled=yes"
 export TERM=xterm
 
-export VISIONOS_SDK="26.0"
+export VISIONOS_SDK="26.5"
```
akien-mga, line 23: Likewise, let's expose this in the build container.

akien-mga:
I'll merge this now to match https://github.com/godotengine/build-containers/pull/163 which I merged earlier, but I'd like to see the outstanding review comments handled in a follow-up pair of PRs to build-containers and godot-build-scripts, so we can clean up further and prepare new build containers for 4.8-dev2.

build-ios/build.sh
akien-mga: Done in https://github.com/godotengine/godot-build-scripts/pull/164

build-ios/build.sh
akien-mga: Done in https://github.com/godotengine/godot-build-scripts/pull/164

build-macos/build.sh
akien-mga: Done in https://github.com/godotengine/godot-build-scripts/pull/164

build-visionos/build.sh
akien-mga: Done in https://github.com/godotengine/godot-build-scripts/pull/164

PR #160, opened 2026-04-24
Extend version validator to check status

Repiteo:
Will prevent future issues where the status passed doesn't match the status in `version.py`

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -118,6 +118,41 @@ case "$choice" in
 esac
 export GODOT_VERSION_STATUS="${status}"
 
+if [ "${skip_git_checkout}" == 0 ]; then
+  git clone https://github.com/godotengine/godot git || /bin/true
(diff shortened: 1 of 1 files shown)


akien-mga:
This should work in usual/official build cases, but I sometimes make test builds which may not fit the pattern, e.g. the `4.7.dotnet[1-3]` builds we've made while status was still `dev`. Or sometimes things like `4.7.beta1+pr1234` when trying to get some testing for a given PR before merging it.

I think we may want to add some interactivity here and flag if there's a status mismatch and give the option to abort or continue anyway.

Repiteo:
Made it prompt on mismatch, as well as moving the validation steps higher up so a user wouldn't have to wait as long for a potential extra input after `sh clean.sh`

akien-mga (approved):

PR #162, opened 2026-06-08
Update Mesa NIR libraries to 25.3.1-3

Repiteo:
Related: godotengine/godot#120025


akien-mga (approved):

akien-mga:
> Depends on [godotengine/godot#120025](https://github.com/godotengine/godot/pull/120025)

Doesn't really depend on it, we don't use that install script for official builds as we pre-install the deps (but you need to delete `deps` after updating this).

PR #166, opened 2026-08-18
visionOS: Generate the classical templates archive

stuartcarnie:
Uncomment the steps to archive the classical visionOS templates.

Depends on:

- https://github.com/godotengine/godot/pull/122559

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -378,17 +378,17 @@ if [ "${build_classical}" == "1" ]; then
 
   ## visionOS (Classical) ##
 
-  #rm -rf visionos_xcode
-  #cp -r git/misc/dist/apple_embedded_xcode visionos_xcode
-  #cp out/visionos/templates/libgodot.visionos.a visionos_xcode/libgodot.visionos.release.xcframework/xros-arm64/libgodot.a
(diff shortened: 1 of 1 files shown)


akien-mga:
What about the Mono flavor? Doesn't it work?

stuartcarnie:
We're not building that yet – as it has not been validated or tested. I admit I've never tried the C# build before 😊

akien-mga:
We are building it currently: https://github.com/godotengine/godot-build-scripts/blob/main/build-visionos/build.sh#L45-L65

But it needs to be tested.

stuartcarnie:
@akien-mga ok, cool – that's good news. If it built, it probably works then, as I imagine it's no different to iOS. I wouldn't know where to start to test it. Presumably we have a C# project I can test.

Incidentally, I am familiar with C#, as I wrote .NET code for over 10 years, but haven't touched C# for over 10 years.

akien-mga:
We have a few C# demos in https://github.com/godotengine/godot-demo-projects/tree/master/mono

But a simple hello world project with `GD.Print("Hello world");` in the defaut Node template should also get you started easily.

Repiteo (approved):
Let it ride

Repiteo:
Thanks!

PR #168, opened 2026-08-20
Backport macOS code signing and notarization logic

Repiteo:
This backports logic from #149, #152, and #154 for 3.x builds. Proved too much for a simple cherry-pick, as references to `osx` were renamed to `macos` in 4.0; these references were reverted back to "osx" where appropriate. I'm unsure if the called scripts specific to the build server account for the "osx" naming scheme, so the changes here should be given a test-drive before fully integrating

### DIFF
--- a/build-release.sh
+++ b/build-release.sh
@@ -26,73 +26,19 @@ sign_windows() {
 }
 
 sign_macos() {
-  if [ -z "${OSX_HOST}" ]; then
-    return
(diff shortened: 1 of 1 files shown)


Repiteo:
Used this for the 3.6.3 release; assuming everything is confirmed to be in working order from macOS users, we'll be good to merge this as-is

PR #171, opened 2026-09-07
publish-release: Upload Android library to MavenCentral

akien-mga:
We missed backporting this part with `publish-release.sh` in the `3.x`/`3.6` branches: https://github.com/godotengine/godot/issues/108161#issuecomment-5484276191

It used to require `sudo` and thus wasn't part of `build-release.sh` at the time, but now we have passwordless `sudo podman` so this should work ok in `publish-release.sh`. To be confirmed with the next 3.7 or 3.6.x build.

To work properly on `3.6`, it requires backporting some gradle changes though, see https://github.com/godotengine/godot/issues/108161#issuecomment-5567411062.

FYI @Repiteo @m4gr3d


akien-mga:
Pushed to `3.6` too (the branch is still in sync with `3.x` for now).

PR #173, opened 2026-09-21
Apple: Update MoltenVK to 1.4.2, fixes Intel/NVIDIA Mac regressions

akien-mga:
Should fix - https://github.com/godotengine/godot/issues/123421.

Needs validating with a build to be sure.

If it works, should be cherry-picked to `4.7` branch (not sure about older branches which used older MoltenVK).


bruvzg:
Other related issues that likely will be fixed by MoltenVK update (can't test, since only some Intel Macs are affected):

https://github.com/godotengine/godot/issues/120897
https://github.com/godotengine/godot/issues/120973
https://github.com/godotengine/godot/issues/121657

Repiteo:
Thanks!

We'll ratify if the latter 3 issues are fixed once dev7 drops

